Issue #9, opened 2019-01-17
"-" or "~" should turn a PDGID object from particle to its anti-particle


HDembinski:
I think the operator logic for the PDGID itself should also apply to its wrapper, the PDGID class. The "-" operator should be defined to turn particles into anti-particles. e.g. -pi_plus == pi_minus. Before jumping ahead and implementing this, a decision has to be made, because the "~" operator could also be used in this way.

* I personally prefer operator "\~", because it conveys the concept of "anti" better. When people write about anti particles in ASCII, I have seen the notation with "~", but never with "-".

* For people who are already used to the logic of raw PDGIDs, using operator "-" is more obvious than "~".

I could live with both decisions.

henryiii:
Particle should and already does use `~`, but PDGID uses as of that PR since it really is just a nicer int.

HDembinski:
So as far as I understand, it is already implemented in #11, then this can be closed.

PR #11, opened 2019-01-17
Adding int version of PDGid

henryiii:
This makes PDGids much simpler; they are just ints with some extra printing and method attached. No need to use `.pgdid` anymore. Comparisons, negation (such as in #9 ) are naturally supported.

This should nicely complement the new Particle class being moved from DecayLanguage.


HDembinski:
I also like this :)

PR #14, opened 2019-01-18
Smaller fixes

henryiii:
Should allow -p to be the correct class. Also changing the meaning of inversion to also behave more like you'd expect for a PDGID.

### DIFF
--- a/particle/pdgid/pdgid.py
+++ b/particle/pdgid/pdgid.py
@@ -19,13 +19,19 @@ class PDGID(int):
     >>> PDGID(11).is_lepton
     True
     """
-    __slots__ = ()
+    __slots__ = () # Keep PDGID a slots based class
 
     def __repr__(self):
         return "<PDGID: {:d}{:s}>".format(int(self),'' if self.is_valid else ' (is_valid==False)')
 
     def __str__(self):
         return repr(self)
+
+    def __neg__(self):
+        return self.__class__(-int(self))
+
+    def __invert__(self):
+        return self.__class__(-int(self))
 
 # Decorate the PDGID class with all relevant functions defined in the pdgid.functions module
 _exclude = ('IntEnum', 'Location', 'print_function', 'division', 'absolute_import')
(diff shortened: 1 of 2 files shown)


particle/pdgid/pdgid.py
```diff
@@ -19,14 +19,20 @@ class PDGID(int):
...
     def __repr__(self):
         return "<PDGID: {:d}{:s}>".format(int(self),'' if self.is_valid else ' (is_valid==False)')
 
     def __str__(self):
         return repr(self)
 
+    def __neg__(self):
+        return self.__class__(-int(self))
+
+    def __invert__(self):
```
eduardo-rodrigues: Why not just return `__neg__` so that it immediately jumps to the eye that the 2 methods are set to do the same?

tests/pdgid/test_pdgid.py
```diff
@@ -20,6 +20,11 @@ def test_class_operations(PDGIDs):
     assert id_positron == - id_electron
     assert PDGIDs.Positron == - id_electron
 
+def test_pdgint():
```
eduardo-rodrigues: What does "pdgint" hint at? You could have added this to the tests on operations that I defined in the previous PR. Not a problem, though, of course.

henryiii: This was written before the previous PR. This should be renamed, probably, but having more sections is a good thing.

particle/pdgid/pdgid.py
henryiii: Yes, I should. I was thinking about allowing discussions on this, since this `~` means something different on ints, so that `~int(x) != ~PDGID(x)`.

tests/pdgid/test_pdgid.py
eduardo-rodrigues: OK, I let you rename and then we're good to go.

henryiii:
I've also added a `print_all` function.

PR #16, opened 2019-01-21
WIP: Adding particles

henryiii:
Still needs tests, adding attr as requirement, and a few more things.


particle/data/particle2008.csv
```diff
@@ -0,0 +1,558 @@
+ID,Mass,MassUpper,MassLower,Width,WidthUpper,WidthLower,I,G,P,C,Anti,Rank,Status,Name,Quarks,Latex
```
eduardo-rodrigues: I would suggest the following for these generated files, as a header:
- Explain which code generated them.
- Specify the input file.

I think this is important.

particle/data/particle2008.csv
```diff
@@ -0,0 +1,558 @@
...
+-9040213,2.149,0.017,0.017,0.36,0.04,0.04,1,1,-1,0,2,0,2,rho(2150),Ud,\rho(2150)^{-}
```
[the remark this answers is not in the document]
henryiii: This should be sorted by increasing PDGID already. Sorting by increasing absID might be nicer, though.

particle/data/particle2008.csv
henryiii: I don't think adding the lines to the file would be easy - to keep the nice and simple csv parser instead of Pandas, I don't think we can have non-csv data in a csv file. However, I can add a README.md in the same directory that explains the generation and upkeep of these files.

particle/data/pdgID_to_latex.txt
```diff
@@ -0,0 +1,356 @@
...
+525,B_{2}^{*+},B_{2}^{*-}
+531,B_s^{0},B_s^{0}
+533,B_s^{*0},B_s^{*0}
+535,B_{s2}^{*0},B_{s2}^{*0}
+541,B_c^{+},B_c^{-}
+551,\eta_b(1S),\eta_b(1S)
+553,\Upsilon(1S),\Upsilon(1S)
+555,\chi_{b2}(1P),\chi_{b2}(1P)
+1112,\Delta(1620)^-,\Delta(1620)^+
+1114,\Delta^-,\Delta^+
```
eduardo-rodrigues: It will be confusing for the reader if the "bar" is not present where needed, e.g. the anti-particle of the Delta- is the Deltabar+. This should not be confused with the Delta+ particle.

I really think the 3rd column should be the full antiparticle name.

particle/particle/__main__.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import absolute_import
+from __future__ import print_function
+
+from .particle import Particle
+
+# This file requires Plumbum
+from plumbum import cli
+
+
+class DecayLanguageParticle(cli.Application):
```
eduardo-rodrigues: To be renamed, I guess ;-).

particle/particle/__main__.py
```diff
@@ -0,0 +1,24 @@
...
+            if value.isnumeric():
+                particle = Particle.from_pdgid(int(value))
+            else:
+                particle = Particle.from_string(value)
+
+            print(particle.describe())
+            print()
+        
+
+DecayLanguageParticle.run()
```
eduardo-rodrigues: Ditto.

particle/particle/convert.py
```diff
@@ -0,0 +1,145 @@
+# This is a conversion file
```
eduardo-rodrigues: A full docstring in triple quotes seems more appropriate, with details on how to use it . simple example for the 2018 CSV file, say.

particle/particle/convert.py
```diff
@@ -0,0 +1,145 @@
...
+    if latexes is None:
+        latexes = (open_text(data, 'pdgID_to_latex.txt'),)
+
+    latex_series = pd.concat([get_from_latex(latex) for latex in latexes])
+    full = full.assign(Latex=latex_series)
+
+    # These items are not very important - can be reconstructed from the PDG
+    del full['Charge'], full['J']
+    
+    # This should be in GeV, not MeV
```
eduardo-rodrigues: I'm confused and the issue of units is a matter where I have very strong feelings: we should return HEP units throughout the Scikit-HEP project, and those are the standard (some do it wrong!). So MeV, nanoseconds and millimeters.

particle/data/particle2008.csv
eduardo-rodrigues: That's for sure an alternative. But it seemed from me (https://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html) that headers were possible.

particle/data/particle2008.csv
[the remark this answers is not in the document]
eduardo-rodrigues: Yep 👍 .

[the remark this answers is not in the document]
henryiii: Done.

particle/data/particle2008.csv
henryiii: Yes, in Pandas. I'm not requiring Pandas, just the standard library csv parser.

particle/particle/__main__.py
henryiii: Was planning to rewrite. Which I did. :)

particle/particle/convert.py
henryiii: +1

particle/data/pdgID_to_latex.txt
henryiii: I think I would rather remove the third column; it's computable from the second column and the extended information.

particle/particle/__main__.py
henryiii: I've settled on:

```bash
python -m particle search 211
```
And, I've added, for your PDGID functionality:

```bash
python -m particle pdgid 211
```

Try it out! (And, I've renamed `pdgid.print_all()` to `print(pdgid.info())`)

particle/particle/convert.py
henryiii: MeV is now used, and it is included in the repr.

eduardo-rodrigues: Thanks!

PR #20, opened 2019-01-23
Fixing issues with reading * in files, ThreeCharge

henryiii:
This makes several changes and fixes some bad bugs:

* `Par` and `Charge` are now separate, as they should be. `Charge` is now internally `3*charge`; if you directly use the charge in a calculation as an int, it will be 3*charge. It's better to compare with the Enums.
* Comment characters are no longer used in the fwf reading. This fixes several bugs in the particle names with missing parts, since `*` is also valid in particle names! Now the commented lines must be removed by hand using `skiprows`. The particle search procedure was simplified since it used to have to work around this bug.
* Smarter uncertainty formatting for large numbers.
* Lower uncertainty values are now positive.
* Programmatic name is slightly more powerful.
* Quarks are now included.

Todo:
- [x] `Par` could be renamed to `Parity`.
- [x]  Rename `charge` to `three_charge`, add `charge` as a numerical shortcut


particle/particle/enums.py
```diff
@@ -43,12 +52,19 @@ class Status(IntEnum):
 
 
 # Mappings that allow the above classes to be produced from text mappings
-Par_mapping = {'+': Par.p, '0': Par.o, '+2/3': Par.u,
-               '++': Par.pp, '-': Par.m, '-1/3': Par.u, '?': Par.u, '': Par.o}
+Par_mapping = {'+': Par.p, '0': Par.o, '-': Par.u, '?': Par.u, '': Par.o}
+Charge_mapping = {'+': Charge.p, '0': Charge.o, '+2/3': Charge.p23,
```
eduardo-rodrigues: Why did you not add entries for -2/3 and +1/3? Unless I missed it.

eduardo-rodrigues:
Yep, Par -> Parity seems a good idea.

Changes seem just fine from a quick check. More difficult is to check the csv files. That can only be done by effectively testing all the code required … I know we have the test suite to be extended and finalised ;-).

particle/particle/enums.py
henryiii: They are not used, but I added them anyway now just in case someone uses these for something else.

henryiii:
Yes, better test suite is high priority! I'm done for a while now.

particle/particle/enums.py
eduardo-rodrigues: Yep, better avoid confusion with not having them ;-). Thanks.

eduardo-rodrigues:
I will help there too, as you know.

eduardo-rodrigues:
Ready to get merged, right?

henryiii:
Almost, found a quick bug

henryiii:
Should be fine now.

PR #23, opened 2019-01-23
Add a requirements file

eduardo-rodrigues:


henryiii:
What about adding a dev requirements file too? Or a Pipfile (which can contain both)?

eduardo-rodrigues:
That's a good idea. But I have so far no experience with that so would not know what to write in the Pipfile as of today … Is that something quick for you? Thanks.

PR #36, opened 2019-01-28
Add a .coveragerc file

eduardo-rodrigues:
This file will make sure that the `particle/particle/convert.py` file is not counted for test coverage, which makes sense.

Other files may be added in the future, of course.


eduardo-rodrigues:
I'm pretty sure this is safe. Let me merge this since you cannot do it just now.

PR #44, opened 2019-01-29
Add a Zenodo DOI tag

eduardo-rodrigues:
For some reason the logo does not appear on PyPI for the 0.2.0 release.
I'm trying to fix that using the full path ...


henryiii:
I noticed that too; I checked DecayLanguage but it has the same issue, so no help there. The full path is probably the correct way to do it.

PR #47, opened 2019-02-02
Adding new, more powerful search

henryiii:
More Pythonic search (the original function was designed around the orignal Pandas table). Some examples of usage:

```python
# These return [p, pbar]
Particle.from_search_list(lambda p: p.name == 'p')
Particle.from_search_list(name = 'p')
Particle.from_search_list(name = lambda x: x == 'p')

# These return 1 result, [p]
Particle.from_search_list(lambda x: x.name == 'p' and x > 0)
Particle.from_search_list(name = 'p', particle=True)

# These return all particles with `p` in the fullname
Particle.from_search_list('p')
Particle.from_search_list(fullname = lambda x: 'p' in x)

# Comparisons, etc.
Particle.from_search_list(lifetime = lambda x : x > 1*second) 
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -127,7 +127,7 @@ can use a search:
     >>> Particle.from_pdgid(211)
     <Particle: pdgid=211, fullname='pi+', mass=139.57061 ± 0.00024 MeV>
     >>>
-    >>> Particle.from_search_list('pi')[0]
+    >>> Particle.from_search_list(lambda p: 'pi' in p.name)[0]
     <Particle: pdgid=111, fullname='pi0', mass=134.9770 ± 0.0005 MeV>
 
 You can search for the properties using keyword arguments, which are
@@ -136,8 +136,10 @@ You can search for the properties using keyword arguments, which are
 ``mass_upper``, ``mass_lower``, ``width_upper``, and ``width_lower``
 (some of those don\'t make sense).
 The alternative ``.from_search()`` requires only one match returned by the search.
-You can also use the first two arguments, called ``name_s`` and ``latex_s``
-to do a loose search, and ``name_re`` and ``latex_re`` to do a regular expression search.
+You can pass a callable or an exact match. You can also build
+the search yourself with the first positional argument, which accepts a callable that is
+given the particle object itself, as in the example above. `particle` can be set to `True`/`False`,
+as well, to limit the search to particles or antiparticles.
 
 Once you have a particle, any of the properties can be accessed, along with several methods.
 Though they are not real properties, you can access ``bar``, ``radius``, and ``spin_type``.
--- a/README.rst
+++ b/README.rst
@@ -118,8 +118,8 @@ You can quickly display PDGID info from the command line with:
 Getting started: Particles
 --------------------------
 
-You can use a variety of methods to get particles. If you know the PDGID number you can get a particle directly, or you
-can use a search:
+You can use a variety of methods to get particles. If you know the PDGID number
+you can get a particle directly, or you can use a search:
 
 .. code-block:: python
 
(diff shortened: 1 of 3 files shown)


eduardo-rodrigues:
I like the idea that one can still do simple things as before, such as `Particle.from_search_list(name = 'p')`, while being exposed to a lot of flexibility and power with for example `Particle.from_search_list(lifetime = lambda x : x > 100000)`.

I did not understand
`Particle.from_search_list(name = 'p', particle=True) # exact match, 1 results`

since the from_search_list is there to return a list of matches.

I look forward to some tests,, which will also serve as examples. Worth updating the README too.

Thanks!

henryiii:
> since the from_search_list is there to return a list of matches.

I'm just adding a note there about what it returns. It's still a list. For that search, it returns a proton. The one(s) above return protons and antiprotons.

henryiii:
I've added tests, fixed an existing bug with the location of `*`, and added better minimum requirements (Requested by @HDembinski). I've dropped the special regular expression handling; I assume any user familiar with regex's could also write a lambda function. This should be good to go unless someone has feedback.

I don't allow unlimited lambda functions with `and` style combining; it's a little uglier to do right in Python 2. I think the current system is versatile enough.

The meaning of an empty search may have changed, and this is an API change for this function.

henryiii:
Note: Are we planning to release a 0.2.1 before 0.3.0? A few of the fixes need to go into 0.2.1 if so, but the API change needs to be only in 0.3.0.

eduardo-rodrigues:
That was the plan, see https://github.com/scikit-hep/particle/milestone/1. But we can go straight for 0.3.0 if that's easiest at this stage. Otherwise, as you say, this PR has things for 0.2.1 and others for 0.3.0, it seems.
What's your preference?

README.rst
```diff
@@ -127,7 +127,7 @@ can use a search:
     >>> Particle.from_pdgid(211)
     <Particle: pdgid=211, fullname='pi+', mass=139.57061 ± 0.00024 MeV>
     >>>
-    >>> Particle.from_search_list('pi')[0]
```
eduardo-rodrigues: By removing this line you mean that this simple shortcut with just a string as name search won't be possible anymore? Not the end of the world but a bit of a shame IMO.

henryiii: This has become `name='pi'`, but that's exact match only. So the proper replacement for this is:

```Python
Particle.from_search_list(name = lambda x: 'pi' in x)[0]
```

Would it make sense to use the first argument as a fuzzy name search if it is not callable? Then this shortcut would return.

eduardo-rodrigues: I'm confused now since from_search_list is different from from_search in that the former returns a list of matches. I guess the answer to your question is yes - if I understand correctly.

eduardo-rodrigues:
Let me know about the idea of getting this all into release 0.2.1 or not … Thanks.

henryiii:
I pulled out the fixes for 0.2.1. We can try to push a release very soon, then merge this toward 0.3.0.

eduardo-rodrigues:
OK, thanks.

README.rst
eduardo-rodrigues: Thanks for that.

eduardo-rodrigues:
I will tag a release 0.2.1 later today unless I hear from you saying to wait …

eduardo-rodrigues:
Version 0.2.1 is tagged and deployed. You can now rebase ...

henryiii:
Any ideas why Azure didn't try to run?

eduardo-rodrigues:
I've noticed Azure do that every now and then but have no idea why it happens. When it happens, I go to the Checks tab and "re-run all".

eduardo-rodrigues:
This is good to go, @henryiii?

eduardo-rodrigues:
FYI I've just released and deployed version 0.2.2 to fix a bug reported by Hans. You will need to pull in the updates after you merge this.

PR #51, opened 2019-02-04
Some 0.2.1 fixes

henryiii:
Some small fixes for 0.2.1.


eduardo-rodrigues:
Good to go!

eduardo-rodrigues:
Let me know if you still want to address the other issue we had set in the 0.2.1 milestone, of if you prefer to postpone. Otherwise we can tag.

henryiii:
Let's make that a 0.3.0 milestone; It's partially completed (Added the most important AmpGen particles), and DecFiles will probably not make it before 0.3.0.

Issue #60, opened 2019-02-20
Parity not loaded in some cases?


apuignav:
Hi,

I'm trying to read the parity for some particles, and even though I can see it in `https://github.com/scikit-hep/particle/blob/master/particle/data/mass_width_2008.fwf`, it's not properly loaded.

```
>>> from particle import Particle
>>> p=Particle.from_pdgid(5122)
>>> print(p.describe())
Name: Lambda(b)  ID: 5122         Fullname: Lambda(b)0     Latex: $\Lambda_b^{0}$
Mass  = 5619.60 ± 0.17 MeV
Lifetime = 1.470e-03 ± 1e-05 ns
I (isospin)       = 0      G (parity)        = 0      Q (charge)       = 0
J (total angular) = 0.5    C (charge parity) = 0      P (space parity) = +
    Quarks: udb
    Antiparticle status: Full (antiparticle name: Lambda(b)~0)
>>> p=Particle.from_pdgid(313)
>>> print(p.describe())
Name: K*(892)    ID: 313          Fullname: K*(892)0       Latex: $K^{*}(892)^{0}$
Mass  = 895.55 ± 0.20 MeV
Width = 47.3 ± 0.5 MeV
I (isospin)       = 1/2    G (parity)        = 0      Q (charge)       = 0
J (total angular) = 1.0    C (charge parity) = 0      P (space parity) = ?
    Quarks: dS
    Antiparticle status: Full (antiparticle name: K*(892)~0)
```

Am I misunderstanding something?

eduardo-rodrigues:
Thanks for reporting, @apuignav.

You are not misunderstanding anything expect for the fact that the file loaded is in fact the digested .csv file ;-). I had a quick look and did find a one-character bug that explains the issue. I will have to regenerate the .csv file … How urgent is your need? I'm a bit tied up right now.

apuignav:
It surely can wait a few days. Thanks a lot!

eduardo-rodrigues:
OK, cool. That helps me 👍 .

I will try and add more tests as this one thing would have been caught had that part of the code been unit tested - 95% coverage "only", so far …
In fact, for this kind of package, there's "code tests" and "physics tests". We have quite a lot of both but for sure more is needed.

Thanks for spotting this.

eduardo-rodrigues:
Closed with https://github.com/scikit-hep/particle/pull/61.

FWYI a release will be done soon once we finalise a couple of other things.

PR #61, opened 2019-02-20
Addresses issue #60

eduardo-rodrigues:
Addresses https://github.com/scikit-hep/particle/issues/60.

The new DB files are still to be regenerated with this fix ...


henryiii:
Looks good, will merge as it's a bugfix.

PR #62, opened 2019-02-27
Adding generation test

henryiii:
This will verify the current version of the particle table matches the input sources so the two remain in sync.


henryiii:
Ready for review. Also dropped the Python3 requirement for running the generation.

PR #70, opened 2019-03-09
Rework for simpler usage of convert, new command line args

henryiii:
Cleanup from DecayLanguage. Docstrings still need update.


particle/particle/convert.py
```diff
@@ -93,6 +108,8 @@ def unmap(mapping):
     pdg_table.set_index('ID', inplace=True)
 
     # Assign the positive values LaTeX names
+    print(pdg_table)
```
eduardo-rodrigues: Debug statements to remove ;-).

Issue #71, opened 2019-03-09
Particle.radius property


eduardo-rodrigues:
It is not clear what this "obscure" property is for, nor what it returns. Documentation needs to be added, providing information on what this PDG data is, and a a reference to details in the PDG.

eduardo-rodrigues:
Similar question for the `Particle.spin_type` property:

it is not clear where this property is taken from. From the PDG? Info should be added on where the definitions of `Particle.spin_type` and `SpinType` enum are taken from.

tevans1260:
The particle radius is a rather hand-wavey number that is used to regulate the behavior of amplitudes at large momentum transfers, based on the classical Q.M arguments in "Theoretical Nuclear Physics" from Blatt and Weisskopf. IMHO this probably shouldn't be included in the basic PDG properties of a particle, as it isn't really a fundamental parameter and isn't really measured for different particles. It's normally arbitrarily set to be between 1 and 2 GeV for light resonances and between 3 and 5 GeV for charm/beauty.

eduardo-rodrigues:
Thanks a lot for the insight, @tevans1260! What about the spin_type and the related enum, see https://github.com/scikit-hep/particle/blob/master/particle/particle/particle.py#L304 and https://github.com/scikit-hep/particle/blob/master/particle/particle/enums.py#L11?

eduardo-rodrigues:
@henryiii, ok with removing the `Particle.radius` property, then? This makes complete sense to me as not a general PDG particle property.

henryiii:
We can, but let's make sure we don't set slots (it's great for developing, but users might want to add arbitrary information). Also, in response to something else somewhere, particle instances are completely editable. If users want to change the mass, etc, they can.

eduardo-rodrigues:
Closed via https://github.com/scikit-hep/particle/pull/85.

PR #73, opened 2019-03-11
Doc and tests for Particle.spin_type

eduardo-rodrigues:
@henryiii, it is not clear where this property is taken from. From the PDG, I guess? I'm here adding doc and tests, extending a bit the function itself, following what seems reasonable to me. Can you point me to the PDG doc where you took `Particle.spin_type` `and `SpinType enum definitions from? Thanks.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -1,41 +1,47 @@
 Changelog
 =========
 
-Version 0.3.1
+Version 0.4.0
 -------------
 Under development
 
-* More package documentation.
+* Rename ``Particle.fullname/name`` to ``Particle.name/pdgname``.
+* More package documentation (data files, ``particle`` submodule).
+* Test suite of ``Particle`` class extended.
+* Rework for simpler usage of ``particle.particle.convert`` (non-public helper module).
 
 Version 0.3.0
 -------------
 March 6th, 2019
 
-* Particle search engine replaced with more intuitive and powerful version.
+* ``Particle`` search engine replaced with more intuitive and powerful version.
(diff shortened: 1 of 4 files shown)


eduardo-rodrigues:
I think this PR can still go in, @henryiii. Then I will proceed and tag a new release.

henryiii:
I'll have to check on this and the other one. I don't think we need to rush a release too fast, I'd like to make sure it works nicely with DecayLanguage first. We can install from git instead of pypi to test.

henryiii:
(I think both these properties came from AmpGen, btw.)

eduardo-rodrigues:
That's a good point. Do not push the merge button if you want to check on SpinType before merging. Else we're good here :-).

henryiii:
I think we can merge. Let's just add the check to issue #71

tevans1260:
I don't think this comes from AmpGen (which in turn, inherits this stuff from MINT). The preferred numbers to store are probably twice the spin (spectroscopy style) and the parity, separately. If you want to combine them to one signed number, that's also fine. Anyway, is this not just taken from the 'J' column of the PDG MC tables + the P -column?

eduardo-rodrigues:
OK, then let's keep `SpinType` is as I can see how to nicely make use of it - trivially find all pseudoscalar charm mesons, for example:
`Particle.from_search_list(lambda p: p.pdgid.is_meson and p.pdgid.has_charm and p.spin_type==SpinType.PseudoScalar)`

henryiii:
Yes, I think it is also part of the printout, and just a computed property.

PR #74, opened 2019-03-11
Rename Particle.bar to Particle.is_name_barred

eduardo-rodrigues:
Another small change to the `Particle` API, @henryiii. Seems better.

All for the imminent release 0.4.0.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,6 +2,7 @@
     :alt: particle
     :target: https://github.com/scikit-hep/particle
 
+
 Particle: PDG particle data and identification codes
 ====================================================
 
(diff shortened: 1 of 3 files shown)


henryiii:
How about is_barred or is_bar? This seems long and a bit too wordy. I do think the "is" is useful, so a user doesn't try to add a bar with it.

eduardo-rodrigues:
It's a bit wordy but having "name" in the method name does say that we're talking about particle names. It's a method code can handily use but I do not see users needing it very often. I would still prefer is_name_barred,so replacing has with is :-).

PR #77, opened 2019-03-12
Finalise tests of pdgid submodule

eduardo-rodrigues:


tests/pdgid/test_functions.py
```diff
@@ -36,6 +36,8 @@
 from particle.pdgid import A
 from particle.pdgid import Z
 
+from ..conftest import PDGIDsEnum
```
henryiii: It seems odd to import from conftest. Maybe you should move that to a general location in the tests folder and import from it in conftest and here?

eduardo-rodrigues: Indeed.

You were too fast. I moved to something more standard. Still not fully happy, but OK.

eduardo-rodrigues:
Now ready for review, @henryiii.

PR #78, opened 2019-03-13
Particle.three_charge made to return an int rather than an Enum

eduardo-rodrigues:


particle/particle/convert.py
```diff
@@ -163,7 +163,8 @@ def unmap(mapping):
     # This will override any negative values
     full.Latex.update(latex_series)
 
-    # These items are not very important - can be reconstructed from the PDG
+    # These items are not very important - can be reconstructed from the PDG ID
+    # TODO: maybe first check the consistency between what is read in and what the PDG ID provides (being maniac)?
```
henryiii: 👍 This was done at one point when this was in a Jupyter notebook, but was removed once it matched. Should be added back in now that this is in a library file.

particle/particle/particle.py
```diff
@@ -264,12 +264,13 @@ def S(self):
...
     @property
     def charge(self):
-       return self.three_charge / 3
+        """The particle charge, in units of the positron charge."""
+        return self.three_charge / 3
 
     @property
     def three_charge(self):
-        'The particle charge (integer * 3).'
-        return Charge(self.pdgid.three_charge)
```
henryiii: Wasn't Charge partially used because it could produce nice output, like "+" for 3? Do we still get that easily? And you can write Charge.p instead of 3.

particle/particle/convert.py
eduardo-rodrigues: This check will be left for post a release. There's already A LOT since 0.3.0 and a release will be necessary for the outstanding PRs in decaylanguage to pass.

henryiii: Yes, I was just agreeing with the comment. We should test before a release by using `python -m pip install git+https://github.com/scikit-hep/particle.git@master` so that we don't risk having an immediate 0.4.1, 0.4.2, etc.

particle/particle/particle.py
eduardo-rodrigues: I don-t know ;-). But I find it non-ideal to get the following:

```
In [3]: Particle.from_pdgid(111).three_charge
Out[3]: <Charge.o: 0>

In [4]: Particle.from_pdgid(-211).three_charge
Out[4]: <Charge.m: -3>
```

If `charge` gives a float, then `three_charge` should give an int, for consistency. Agree?

henryiii: The IntEnum behaves exactly like an int under duck typing, though.

eduardo-rodrigues: Sure, sure. What annoys me is that the user will get `Charge.m: -3>` and alike. I would prefer to have `Particle.three_charge` consistent with `PDGID.three_charge`, so plain ints. The Enums are really nice for all the conversion and handling machinery you wrote, and work perfectly there.

henryiii: By my same argument, `Particle.from_pdgid(111).three_charge == Charge.o` will work regardless of which one it returns. So I would be okay with it returning a plain int, as I think that's the main reason to have the enum.

PR #80, opened 2019-03-13
Extension of tests & doc for Particle.from_XXX methods

eduardo-rodrigues:


henryiii:
Nice to see progress here!

particle/particle/particle.py
```diff
@@ -452,13 +452,13 @@ def from_search_list(cls, filter_fn=None, particle=None, **search_terms):
...
-        candidate, and returns true/false. Example:
+        candidate, and returns True/False. Example:
 
             >>> Particle.from_search_list(lambda p: 'p' in p.name)
             # Returns list of all particles with p somewhere in name
 
-        You can also pass particle=True/False to force a particle or antiparticle. If
-        this is not callable, it will do a "fuzzy" search on the name. So this is identical:
+        You can pass particle=True/False to force a particle or antiparticle.
+        IF this is not callable, it will do a "fuzzy" search on the name. So this is identical:
```
henryiii: Why IF instead of If?

eduardo-rodrigues:
Thanks - you can see I've spent quite some time on this package today ;-). It is indeed nice to see the package being streamlined.

particle/particle/particle.py
eduardo-rodrigues: Good spot - Atom played my tricks :S. Of course it should be "If" ... will fix ...

eduardo-rodrigues: Done.

eduardo-rodrigues:
I think I'm done here, @henryiii. Anything you may suggest? Else I will merge. Thanks again.

Issue #81, opened 2019-03-13
Add tests for Particle.html_name


eduardo-rodrigues:
Not a priority but worth writing down that some basic tests should be added.

eduardo-rodrigues:
Done with https://github.com/scikit-hep/particle/pull/83.

PR #83, opened 2019-03-13
Improve handling of, and tests for, particle HTML and LaTeX names

eduardo-rodrigues:
First stab at testing the HTML and LaTeX rendering of particle names. Along the way I made LaTex names more uniform. I also want to improve the HTML names themselves. To be continued ...


eduardo-rodrigues:
First point to discuss, @henryiii:
the code presently renders the HTML names of antiparticles according to the following examples:
```
'~u' : u antiquark
'~\\nu<SUB>\\mu</SUB>') : nu_mu_bar
```
It would make more sense to have the overline rendered in a standard way, say with
`<SPAN STYLE="text-decoration:overline">u</SPAN>`
What's your opinion?

In fact, what are you presently using HTML names in? In notebooks?

eduardo-rodrigues:
Second point for discussion:
At the moment we have a kind of mixed way of rendering greek letters in the HMTL names because only pi, rho and omega are replaced with proper greek letters. Here are some examples:
```
'\\mu<SUP>+</SUP>': mu+
'π<SUP>0</SUP>' : pi0
'π<SUP>-</SUP>' :  pi-
```
The relevant code in Particle.html_name is
`name = name.replace(r'\pi', 'π').replace(r'\rho', 'ρ').replace(r'\omega', 'ω')`

I reckon we should either handle all greek letters, or none. Again, what's your opinion?

A neat way would be to have a little helper function doing all the replacements using the `unicodedata` module with greek letters being defined as for example
```
unicodedata.lookup('GREEK SMALL LETTER ALPHA')
unicodedata.lookup('GREEK CAPITAL LETTER UPSILON')
```
Unless you know of a module to do this "automatically"? The kind of lines above provide an easy way to do all replacements, though.

Thanks in advance for your opinions, @henryiii.

henryiii:
> In fact, what are you presently using HTML names in? In notebooks?

The dot generator accepts html, which converts to svg (and later can be displayed in a notebook). I was thinking about overlines, but don't remember if I tested them or not.

eduardo-rodrigues:
Third point - French say "there's never 2 without 3" ;-):
the `Particle` class has the properties `name`, `pdgname`, `html_name` and `latex`. How about making these uniform with `htmlname` and `latexname` ?

henryiii:
> `unicodedata` module

Wow, didn't know about that one. There still are modules even in the Python std. lib that surprise me. That actually would be easy, I think, let me throw together a PR. Yes, we should do all letters.

eduardo-rodrigues:
This PR addresses https://github.com/scikit-hep/particle/issues/81 (but the tests of HTML names aren't yet comprehensive).

eduardo-rodrigues:
> unicodedata module
> 
> Wow, didn't know about that one. There still are modules even in the Python std. lib that surprise me. That actually would be easy, I think, let me throw together a PR. Yes, we should do all letters.

Don't worry about a new PR. I will put it all in this one since part of its scope ;-). Indeed it's just a matter of a few lines of code - cool.

henryiii:
Okay, you can simply run it on all latex commands, then try to run `unicodedata.lookup('GREEK SMALL LETTER '+ latexcom.upper()) `, and if it works, that's the conversion to unicode. :) Giving it a nice name and location just in case it's useful somewhere else, and a couple of unit tests.

henryiii:
You need to write `[skip ci]` to get it to skip. I like doing this:

```
git commit -m "Something

[skip ci]
```

henryiii:
And note that you need to remove the [skip ci]'s from the squashed message when you squash and merge.

eduardo-rodrigues:
> You need to write [skip ci] to get it to skip. I like doing this:
> git commit -m "Something
> 
> [skip ci]"

I've tried this on my Windows terminal but it didn't work as expected. I'll give it another try soon ...

eduardo-rodrigues:
> And note that you need to remove the [skip ci]'s from the squashed message when you squash and merge.

Sure, I will do a little bit of clean-up in the end ...

eduardo-rodrigues:
> Third point - French say "there's never 2 without 3" ;-):
> the Particle class has the properties name, pdgname, html_name and latex. How about making these uniform with htmlname and latexname ?

OK, done with this :-).

eduardo-rodrigues:
> Okay, you can simply run it on all latex commands, then try to run unicodedata.lookup('GREEK SMALL LETTER '+ latexcom.upper()) , and if it works, that's the conversion to unicode. :) Giving it a nice name and location just in case it's useful somewhere else, and a couple of unit tests.

Turns out it was a bit more complicated, for a couple of reasons. Anyway, all is working :)! I will add yet more tests later … Looking forward to your feedback on this.

eduardo-rodrigues:
BTW, @henryiii, I'm now also using `<SPAN STYLE="text-decoration:overline">u</SPAN>`, cf. my comment earlier.

Oh, I reckon I'm going to take the LaTeX to HTML function out of the class as that's a neat function to have in the package as utilities!

henryiii:
> take the LaTeX to HTML function out of the class

👍

eduardo-rodrigues:
OK, noy fully ready for review! (Side note: I will add a few more tests of particle HTML names before merging.)

docs/CHANGELOG.rst
```diff
@@ -7,11 +7,13 @@ Under development
...
   - Rename ``Particle.fullname/name`` to ``Particle.name/pdgname``.
-  - Rename ``Particle.bar`` to ``Particle.has_name_barred``.
+  - Rename ``Particle.bar`` to ``Particle.is_name_barred``.
+  - Rename ``Particle.latex`` to ``Particle.latexname``.
+  - Rename ``Particle.html_name`` to ``Particle.htmlname``.
 * Neutrinos added to the 2018 data files.
-* Test suite of ``Particle`` class extended.
-* More package documentation (data files, ``particle`` submodule).
-* Rework for simpler usage of ``particle.particle.convert`` (non-public helper module).
+* Test suite of ``particle`` and ``pdgid`` submodules much extended.
```
henryiii:
```suggestion
* Test suite of ``particle`` and ``pdgid`` submodules extended.
```

particle/particle/utilities.py
```diff
@@ -64,3 +65,74 @@ def str_with_unc(value, upper, lower=None):
...
+
+
+def latex_to_html_name(name):
+    """Conversion of particle names from LaTeX to HTML."""
+    name = re.sub(r'\^\{(.*?)\}', r'<SUP>\1</SUP>', name)
+    name = re.sub(r'\_\{(.*?)\}', r'<SUB>\1</SUB>', name)
+    name = re.sub(r'\\mathrm\{(.*?)\}', r'\1', name)
+    name = re.sub(r'\\left\[(.*?)\\right\]', r'[\1] ', name)
+    for gl in list_name_greek_letters:
+        # Special care - unicodedata library uses "lamda" for "lambda" :S!
```
henryiii: Couldn't you first replace lambda with lamda, then try to convert to unicode name? It would be nice to avoid the explicit listing.

particle/particle/utilities.py
```diff
@@ -64,3 +65,74 @@ def str_with_unc(value, upper, lower=None):
...
+    'phi',
+    'pi',
+    'psi',
+    'rho',
+    'sigma',
+    'tau',
+    'theta',
+    'upsilon',
+    'xi',
+    'zeta'
```
henryiii: Can't you list only one of the two caps versions, then use titlecase or lower to make the other half?

eduardo-rodrigues: That won't work because unicodedata.lookup needs to know whether the greek letter name is for a capital letter or not, see the snippet
```
    if letter == letter.lower():
        return unicodedata.lookup('GREEK SMALL LETTER %s' % letter.upper())
    else:
        return unicodedata.lookup('GREEK CAPITAL LETTER %s' % letter.upper())
```
If I give a lowercase letter I would need a Boolean flag ...

Unless you just mean creating the second part of the list `list_name_greek_letters` on the fly?

particle/particle/utilities.py
eduardo-rodrigues: You mean replacing
```
...
name = name.replace(r'\%s'%gl, greek_letter_name_to_unicode(gl.replace('ambda', 'amda')))
```
with
```
... 
name = name.replace('ambda', 'amda')
...
name = name.replace(r'\%s'%gl, greek_letter_name_to_unicode(gl))
```
?

particle/particle/utilities.py
henryiii: I did mean that, but this would work:

```python
return unicodedata.lookup('GREEK {case} LETTER {name}'.format(
    case = 'LOWER' if letter == letter.lower() else 'UPPER',
    name = letter.upper()))
```

PS: Please use `.format` instead of `%`.

particle/particle/utilities.py
henryiii: Yes, esp. if that allows you to avoid the list and just use a try catch block.

henryiii: Something like:

```python
try:
    unicodedata.lookup('GREEK SMALL LETTER MATHRM')
except KeyError:
    return unchanged
```

will return `unchanged`.

particle/particle/utilities.py
eduardo-rodrigues: I will commit this nice short return statement you suggest above.

particle/particle/utilities.py
eduardo-rodrigues: I'm not sure about the above. In any case I've just committed a simplified version of the LaTeX -> HTML conversion code.

eduardo-rodrigues:
Ready to rock!

eduardo-rodrigues:
Not sure - did you still have suggestions, @henryiii, or shall I go ahead and merge this?

henryiii:
Not quite, let me try something.

eduardo-rodrigues:
BTW, I checked some of the complicated HTML names I wrote for the tests and they display just fine.

henryiii:
Verify my change is okay, then it's ready.

particle/particle/utilities.py
henryiii: This would have been done a different way; you'd look for all `/<letters>` and then try to convert them. It's a larger change; let's go with this for now.

Issue #41, opened 2019-01-29
Add tests for DecayLanguage particles


henryiii:
The particles used by DecayLauage should be tested. This would require two tests:

- [x] One that reads in AmpGen style names and makes sure they all convert correctly to Particles.
- [x] One that reads in DecFile style names and makes sure they all convert correctly to Particles.

In both cases, the initial input to the mark.parameterize should be a dump from DecayLanguage of all names detected.

eduardo-rodrigues:
FYI, @henryiii, I started addressing this task in https://github.com/scikit-hep/particle/pull/80.

eduardo-rodrigues:
The search of particles given a DecFile style of name requires some work on the regex. I tried playing a bit and it is not trivial for the regex (`getrec` in `particle.regex` submodule) to properly parse and match all particles, for example the following set
```
anti-K*0
anti-nu_mu
Lambda_b0
Lambda_c+
phi
chi_c0
chi_b1(2P)
```
(Name examples taken from https://github.com/scikit-hep/decaylanguage/blob/master/decaylanguage/data/DECAY_LHCB.DEC.)

eduardo-rodrigues:
Dealt with via https://github.com/scikit-hep/particle/pull/101.

PR #84, opened 2019-03-15
Updates for copyright and licensing

eduardo-rodrigues:


eduardo-rodrigues:
Totally unrelated: I also committed trivial updates to please PEP8 rules.

eduardo-rodrigues:
BTW, to me this is the last commit before releasing 0.4.0. Indeed there are very many important updates and fixes since 0.3.0 and the open "issues" are not a massive concern.

(For completeness: issue https://github.com/scikit-hep/particle/issues/41 will require some work. I will add a few comments there shortly.)

henryiii:
Please do not release till after I get off the train, I’m homing to add some tests on the train.

eduardo-rodrigues:
No problem. There's a reason why I'm sending advance warning ;-). I'm just saying that a release should be imminent … I will of course wait for your green light.

LICENSE
```diff
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2018, The Scikit-HEP Project
+Copyright (c) 2018-2019, The particle developers.
```
henryiii: I think The SciKit-HEP developers would be fine. (And Particle should be capitalized if used)

eduardo-rodrigues: That's "a bit" too much to give everyone contributing to any Scikit-HEP org package the copyright hold, I find. I will capitalise "particle", sure.

eduardo-rodrigues: Following the comments from Hans for boost-histogram, would you prefer to copyright to us two only, just as there? That's OK for such a well-defined package/project done 100% by us anyway.

eduardo-rodrigues:
Shall I then change the copyright to

> Copyright (c) 2018-2019, Eduardo Rodrigues and Henry Schreiner.

or leave it as it is in this PR?

LICENSE
henryiii: I believe we can just put our names here. I have permission to do so, anyway.

eduardo-rodrigues: OK, will do.

eduardo-rodrigues:
OK, @henryiii, I adapted following what Hans and you did for boost-histogram. Ready to go if you're happy too.

PR #86, opened 2019-03-15
Adding list of particles from the LHCb decfile

henryiii:
Here's the list of currently failing particles. It's also rather slow looking up particles, it might be optimizable. Some of these are special and don't need to be in Particle.

```

	dec_names_undealt_with = (
"B'_1+",
 "B'_1-",
 "B'_10",
 "B'_c1+",
 "B'_c1-",
 "B'_s10",
 'B_0*+',
 'B_0*-',
 'B_0*0',
 'B_1+',
 'B_1-',
 'B_10',
 'B_c*+',
 'B_c*-',
 'B_c0*+',
 'B_c0*-',
 'B_c1+',
 'B_c1-',
 'B_c2*+',
 'B_c2*-',
 'B_s10',
 'D(2S)+',
 'D(2S)-',
 'D(2S)0',
 'D*(2S)+',
 'D*(2S)-',
 'D*(2S)0',
 'K_L0',
 'K_S0',
 'Omega_c*0',
 'Sigma_c*+',
 'Sigma_c*++',
 'Sigma_c*0',
 'Upsilon_2(1D)',
 'Xi_c*+',
 'Xi_c*0',
 "anti-B'_10",
 "anti-B'_s10",
 'anti-B_0*0',
 'anti-B_10',
 'anti-B_s10',
 'anti-D(2S)0',
 'anti-D*(2S)0',
 'anti-Sigma_c*-',
 'anti-Sigma_c*--',
 'anti-Sigma_c*0',
 'chi_b0(2P)',
 'chi_b0(3P)',
 'chi_b1(2P)',
 'chi_b1(3P)',
 'chi_b2(2P)',
 'chi_b2(3P)',
 'eta_b2(1D)',
 'eta_b2(2D)',
 'eta_c(2S)',
 'omega(2S)',
 'rho(2S)+',
 'rho(2S)-',
 'rho(2S)0',
 'rho(3S)+',
 'rho(3S)-',
 'rho(3S)0')
```


eduardo-rodrigues:
OK, @henryiii, I had a proper look at the get_dec(name) function. The one thing to do first is to check whether a simpler call get_search(name) returns the one match. That helps a lot and finds many of the particles the now-old code was not finding. I've just committed this.

eduardo-rodrigues:
In a few cases the *only* way out is to resort to a mapping, I'm afraid. Take cases such as "phi" in dec files. Here one needs to specify the phi without "(mass)" is the phi(1020). There are other such cases ...

eduardo-rodrigues:
Oops, I had to revert back a commit as the list you added to test_decfilenames.py is the _full_ list of particles found in .dec, not the ones failing. Sorry.

eduardo-rodrigues:
At least the few lines of code I added to from_dec(...) are already an optimisation of the previous code since the slow regex business will not even be called in many cases ;-).

henryiii:
Feel free to edit the above description with the new list of failing particles. `-VV` to print them out when running pytest.

eduardo-rodrigues:
I put in some changes so that the test passes, having an explicit list of what .dec names aren't dealt with by the present code.

This being said, there are many .dec names for which the returned name with from_dec() is in fact not the right one. In other words, the matching code needs further development. Examples:
```
In [2]: Particle.from_dec('Delta+')
Out[2]: <Particle: pdgid=-1112, name='Delta(1620)~+', mass=1610 ± 20 MeV>
In [2]: Particle.from_dec('Delta+')
Out[2]: <Particle: pdgid=-1112, name='Delta(1620)~+', mass=1610 ± 20 MeV>
In [2]: Particle.from_dec('Delta+')
Out[2]: <Particle: pdgid=-1112, name='Delta(1620)~+', mass=1610 ± 20 MeV>
```
These issues happen because .dec particle names do not really follow a good logic ;-).

eduardo-rodrigues:
A better way to code from_dec() may be addressing in turns specific particle families: neutrinos, quarkonia, etc.

eduardo-rodrigues:
The list of undealt with particles got much smaller :-). This being said, I would not like to keep all the if statements I had to add to Particle.get_dec. That stuff is definitely temporary code, and a better solution should be tried ...

eduardo-rodrigues:
@henryiii, I would suggest I go ahead and merge this PR at this stage. We then come back to the task after a release. This makes sense to me now. We can certainly add to the README file a note saying that `Particle.from_dec` is work in progress.

henryiii:
Rebased and fixed (normal, not dec-file) searches for `b~`.

eduardo-rodrigues:
Just a tiny question. Otherwise looks good to me.

henryiii:
@eduardo-rodrigues What question?

eduardo-rodrigues:
Oh. Just the question above concerning the indentation of the return statement in
```
            if "anti-" in name: 
                name = name.replace('anti-', '') 
                particle = False 
            return cls.find(pdg_name=name, particle=particle)
```

particle/particle/particle.py
```diff
@@ -552,12 +552,69 @@ def find(cls, *args, **search_terms):
...
         mat = getdec.match(name)
+
         if mat is None:
-            return cls.find(name=name)
+            # Deal with antiquarks
+            particle = None
+            if "anti-" in name:
+                name = name.replace('anti-', '')
+                particle = False
+            return cls.find(pdg_name=name, particle=particle)
```
henryiii: We could add a blank line here if that helps, but yes, this is supposed to be outside the if.

eduardo-rodrigues:
A blank line seems good to me, to ease the reading. Thanks.

Issue #87, opened 2019-03-15
Suggestion to change the way particles are printed out


eduardo-rodrigues:
It seems more natural to have
`<Particle: name='Lambda(b)0', pdgid=5122, mass=5619.60 ± 0.17 MeV>`
instead of
`<Particle: pdgid=5122, name='Lambda(b)0', mass=5619.60 ± 0.17 MeV>`

Likewise, the `.describe()` method presently returns
```
PDG name: K          ID: 311          Name: K0             Latex: $K^{0}$
Mass  = 497.611 ± 0.013 MeV
Width = -1.0 MeV
I (isospin)       = 1/2    G (parity)        = 0      Q (charge)       = 0
J (total angular) = 0.0    C (charge parity) = 0      P (space parity) = ?
    Quarks: dS
    Antiparticle status: Full (antiparticle name: K~0)
```

whereas the following seems more nicely organised:
```
Name: K0             ID: 311          Latex: $K^{0}$
Mass  = 497.611 ± 0.013 MeV
Width = -1.0 MeV
Q (charge)       = 0       J (total angular) = 0.0    P (space parity) = ?
C (charge parity) = 0      I (isospin)       = 1/2    G (parity)        = 0
    Quarks: dS
    Antiparticle name: K~0 (antiparticle status: Full)
```
In particular, the internal PDG table name `pdgname` is not printed.
Also the first line provides the most important and relevant quantum numbers. Indeed, C, I and G are not definable for many particles.

Agreed?

henryiii:
I like it better 👍 (with aligned equals though)

eduardo-rodrigues:
Sure - my copy-and-paste was not perfect. But, yes, meant that for sure :-).

eduardo-rodrigues:
Closed via https://github.com/scikit-hep/particle/pull/90.

PR #90, opened 2019-03-15
Better print-out of Particle properties

eduardo-rodrigues:
Addresses https://github.com/scikit-hep/particle/issues/87.


eduardo-rodrigues:
Thanks  @henryiii.

PR #95, opened 2019-03-20
Improvement to Enum defining what happens when a particle is inverted

eduardo-rodrigues:
Seem more natural names for what they reflect.


henryiii:
FYI: I believe the original names were mimicking the original file which had S, F, and B as the letters. I think there is no reason to stick to those names, though, and the new ones are a little nicer.

eduardo-rodrigues:
Indeed, and agreed.

PR #96, opened 2019-03-20
Fix for a couple of particles not dealt with properly in 2008 PDG extended file

eduardo-rodrigues:
One test will fail and I'm not yet understanding why - as if one property ("anti_flag") was not being updated?


particle/data/mass_width_2008_ext.fwf
```diff
@@ -30,6 +30,11 @@
 * Xi(c)(2815)         104322, 104312
 * Xi(c)(2790)         104324, 104314
 *
+* 4) Particles not well-known until recently and entries in the 2018 extended file is not ideal:
```
henryiii:
```suggestion
* 4) Particles not well-known until recently and entries in the 2018 extended file are not ideal:
```

PR #101, opened 2019-03-27
Deal with .dec particles

eduardo-rodrigues:
Addresses https://github.com/scikit-hep/particle/issues/41, getting to the bottom of it … well, at least for now as the list of .dec file specific particles and those not dealt with by the 2018 PDG table will change.


eduardo-rodrigues:
FYI, @henryiii. This is what we can do for now to deal with .dec particle names given the presently known particles in the 2018 PDG table. This is necessary to continue the developments of DecayLanguage.

At some point we may want to replicate the knowledge/list of .dec specific particles to DecayLanguage.

I will merge this tomorrow unless I hear from you.

particle/particle/particle.py
```diff
@@ -564,27 +564,44 @@ def from_dec(cls, name):
...
+            'Upsilon_2(1D)': 'Upsilon(2)(1D)',
             'Upsilon(5S)' : 'Upsilon(10860)',
             'n0': 'n',
             'p+': 'p',
-            'X_1(3872)': 'X(3872)'
+            'X_1(3872)': 'X(3872)',
+            'Omega_c*0': 'Omega(c)(2770)'
         }
-        if name in dec_to_pdg_mapping.keys():
+        if name in sorted(dec_to_pdg_mapping.keys(), reverse=True):
```
henryiii: Why is this sorting? The `in` operator doesn't care, and the time to sort is high compared to checking membership. And once it's sorted, the `in` operator now has to check membership in a list, which is slower than checking membership in an unordered structure like a set (Python < 3.5).

particle/particle/particle.py
```diff
@@ -564,27 +564,44 @@ def from_dec(cls, name):
...
+            'anti-B_1': 'B(1)(5721)~',
             'Sigma*': 'Sigma(1385)',
-            'Xi*': 'Xi(1530)'
+            'Xi*': 'Xi(1530)',
+            'Sigma_c*': 'Sigma(c)(2520)',
+            'Xi_c*': 'Xi(c)(2645)',
+            'B_1': 'B(1)(5721)',
         }
-        for oldw, neww in dec_to_pdg_replacements.items():
+        for oldw, neww in sorted(dec_to_pdg_replacements.items(), reverse=True):
```
henryiii: This one can be affected by the sorting; while you can control the order explicitly in Python 3.6+ by writing the dictionary differently, this is needed in Python 3.5 and below. A comment here about why this needs to be sorted would be a good idea, though. Also, this might be pre-computable to avoid the sort each time.

particle/particle/particle.py
eduardo-rodrigues: Oh, I meant to add a comment next to the code, actually. These sorts are needed to make sure certain replacements are not done twice for names such as "X" and "anti-X".
Agreed that the sorting should be done outside.

particle/particle/particle.py
eduardo-rodrigues: Agreed 100%. See reply above.

PR #103, opened 2019-03-29
Charge is an entry of CSV files again 

eduardo-rodrigues:
Presently, the particle charge was retrieved via the PDGID as Particle.pdgid.charge. This is not good enough if we want - we do - to deal with user-defined particles as in decay files (see needs for the DecayLanguage package). This PR puts charge back as an entry in the particle data CSV files.

Along the way I fixed what seems either a nasty swap of IDs for Xi_b baryons or a bug from the PDG. In any case I added a consistency check between the charge returned from the PDG tables and that from the encoded PDGID (this is how I found the bug, actually).


particle/particle/particle.py
```diff
@@ -219,6 +218,7 @@ def load_table(cls, filename=None, append=False):
                     P=int(v['P']),
                     C=int(v['C']),
                     anti_flag=int(v['Anti']),
+                    Particle_enum_three_charge=int(v['Charge']),
```
henryiii: How about calling this `_threecharge`? This will make the constructor much nicer and will play nicely with "three_charge" being the public-facing version of this.

eduardo-rodrigues: OK. done.

Issue #115, opened 2019-05-08
Bug in the extended PDG file


eduardo-rodrigues:
The extended PDG file, which we use to provide more particle info than that provided in the standard downloadable .mcd files, specify for example how to get the antiparticle name from the particle. There are 3 possibilities: self-conjugate (gamma, pi0, eta), same name but change of charge ( e- <--> e+, pi+ <--> pi-, K+ <--> K-, W+ <--> W- ), or change of charge sign with a bar (nu(e) <--> nubar(e), p <--> pbar, Delta++ <--> Deltabar--).

It turns out that all excited charged B mesons have been described in the 3rd category whereas they fall under the 2nd one, just as B+, D+, Ds+, Bc+. So one is getting for example

> B*+ <--> B*~-

whereas the correct "relation" is

> B*+ <--> B*-

I suggest we correct these in our extension to the extended file, just as we did for other issues, see the comments at the top of that file. I will submit a PR soon.

PR #116, opened 2019-05-08
Fix for bug in the PDG extended file from 2008

eduardo-rodrigues:
Fixes https://github.com/scikit-hep/particle/issues/115.


eduardo-rodrigues:
I will need to make similar changes for the K and D mesons - clearly the PDG must have changed conventions at some point, as that many "bugs" seems unlikely.

eduardo-rodrigues:
OK, ready for review/merge, @henryiii.

eduardo-rodrigues:
Closes https://github.com/scikit-hep/particle/issues/115.

PR #119, opened 2019-05-09
Yet more corner cases of .dec particle names

eduardo-rodrigues:
I found a couple more particles whose .dec particle names are not sufficiently well defined. They are dealt with in a special way, just as for similar cases.

I added tests for the whole family.


henryiii:
This looks fine.

PR #123, opened 2019-05-12
Adding (broken) test for lifetime search

henryiii:
@eduardo-rodrigues , can you fix this?


eduardo-rodrigues:
I've addressed this corner case in PR https://github.com/scikit-hep/particle/pull/128. Let's rerun the CI once the latter is merged.

Issue #118, opened 2019-05-09
Put particle changes in 2008 file


henryiii:
The "ext" file for the 2008 particle file is becoming unwieldy. The changes should be made directly on the 2008 file, with a comment at the top indicating it is a modified file, and with a link to the original. The extension mechanism can still be available for adding new particles.

eduardo-rodrigues:
Quick word to say that I'm working on a new "master data file" with structure in it as well. So this is not forgotten and in fact being dealt with.

PR #137, opened 2019-05-21
PDG particle data table 2019 update

eduardo-rodrigues:
The 2019 update is out!

I need to update our CSV files with this and make 2019 the default …

This should go out in a 0.5.0 release, which should at least include the update described in https://github.com/scikit-hep/particle/issues/118.


eduardo-rodrigues:
Hi @henryiii, this PR is now ready for review. All ready apart from one little thing: I put in a temporary fix in `particle/particle/convert.py` to generate the 2019 CSV file with updates, as otherwise `Pandas `would give me an error. I suspect that's because, as it is, one is effectively adding twice the file with PDG ID - LaTeX name matches. Maybe you have your preferred way to make the user LaTeX file optional when passed to the `convert(...)` function (my temporary fix is just a hack). Thanks.

eduardo-rodrigues:
OK, @henryiii, I've now fixed `particle.particle.convert.py` making the extra LaTeX file optional.

This is ready to go in as far as I'm concerned.

particle/particle/convert.py
```diff
@@ -314,10 +314,11 @@ def main(year):
     produce_files(particle2008, particlenew, year)
 
 
-def convert(fwf, latex, output):
-    table = get_from_pdg_extended(fwf,
-                                  [data.open_text(data, 'pdgid_to_latex.csv'),
-                                   latex])
+def convert(output, fwf, latex):
```
henryiii: Since you support None now, this could be optional (as you said).
```suggestion
def convert(output, fwf, latex=None):
```

eduardo-rodrigues: And I meant to add that. I clearly forgot in the end. Thanks for spotting it.

Issue #140, opened 2019-06-06
Better dealing of undefined/non-measured particle widths


eduardo-rodrigues:
Several known particles have no measured value of their width, e.g. the D∗(2007)0, with a width < 2.1 MeV @ 90% confidence level (see PDGLive). The present code shows
```
>>> p = Particle.from_pdgid(423)
>>> print(p.width, p.width_lower, p.width_upper)
2.1 1.0 -1.0
```
, which is "digested information" from the PDG particle data table: the .mcd file contains empty fields for the 3 quantities whereas the 2018 extended file flags the experimental upper limit with the values `2.1,-1,-1` (the fact the width_lower is +1 in the print out above is because all errors are converted to positive numbers in our converter code that produces the CSV file we use).    

We need to improve the above on 2 fronts:

1) `print(p.width, p.width_lower, p.width_upper)` should rather return `2.1 -1.0 -1.0` for this kind of cases, so that the user can see the errors are not known. The documentation could and should mention this special case.
Some code such as `df.loc[df['WidthLower'] != -1, 'WidthLower'] = 1` could do the trick.

2) With the special `-1.0` values for the width errors the method `Particle.describe()`  could catch these and print the particle width as `Width < 2.1 MeV` whereas now it prints `Width = 2.1 ± 1.0 MeV`, which is wrong and misleading.

One should also look at the mass and mass error entries to catch similar issues.

eduardo-rodrigues:
Better do this only once PRs https://github.com/scikit-hep/particle/pull/137 and https://github.com/scikit-hep/particle/pull/139 are merged.

eduardo-rodrigues:
Closed via https://github.com/scikit-hep/particle/pull/142.

PR #142, opened 2019-06-10
Correctly deal with experimental width upper limits

eduardo-rodrigues:
Addresses https://github.com/scikit-hep/particle/issues/140.


eduardo-rodrigues:
Closes https://github.com/scikit-hep/particle/issues/140.

Issue #141, opened 2019-06-07
Particles only in so-called extended file should not be in the CSV


eduardo-rodrigues:
I noticed that some non-established particles such as the K(4)(2500)+ are present in the 2008 extended file but aren't in the PDG .mcd files. At present these particles make it to our CSV files, so are available to the user. I reckon it is preferable to not have them in the CSV files.
Agreed, @henryiii?

Deal with this once https://github.com/scikit-hep/particle/pull/137 is merged.

eduardo-rodrigues:
@henryiii, do you have a suggestion for how to best perform this filtering?

eduardo-rodrigues:
I have the code to do this and will commit a PR once https://github.com/scikit-hep/particle/pull/142 is merged.

eduardo-rodrigues:
Closed via https://github.com/scikit-hep/particle/pull/144.

PR #144, opened 2019-06-13
Particles in extended file but not in standard PDG file not made available to users

eduardo-rodrigues:
Addresses https://github.com/scikit-hep/particle/issues/141.


eduardo-rodrigues:
This seems the correct behaviour given the PDG standard file with established particles.

particle/particle/convert.py
```diff
@@ -272,6 +274,9 @@ def update_from_mcd(full_table, update_table):
     update_table_neg.index = -update_table_neg.index
     full_table.update(update_table_neg)
 
+    # Only keep rows present in the update table, i.e. the .mcd file
+    full_table = full_table[full_table.index.isin(update_table.index) | full_table.index.isin(update_table_neg.index)]
```
henryiii: This looks fine for now.

eduardo-rodrigues: Yeah, otherwise we were providing particles with semi-OK properties more than a decade old, sometimes with best guesses. That's not great, I reckon. Better just not provide those particles.

docs/CHANGELOG.rst
```diff
@@ -5,15 +5,15 @@ Version 0.5.0
 -------------
 In development.
 
+* Added the 2019 PDG data table, now default.
```
henryiii: Might be worth adding:

```
* Some poorly established particles not in the current PDG data files were previously erroneously made available. They have now been removed.
```

eduardo-rodrigues: Yep. I will add that tomorrow. Thanks for the comments.

Issue #108, opened 2019-04-06
Should we do some doctest tests?


eduardo-rodrigues:
There's a fair amount of examples in the code and it would be ideal to make sure that those examples stay valid at all times. This can often be achieved with doctest tests (some examples may not be full, e.g. because of complex inputs, and those can be skipped).

shantanu-gontia:
The comments in the output will need to be moved elsewhere, otherwise `doctest` wouldn't work. I tried this with the examples in the `kinematics.py` file. 
~~~~
>>> lifetime_to_width(0.001520119980246514*ns)
    4.33  # result again returned in MeV
~~~~ 
fails while
~~~~
>>> lifetime_to_width(0.001520119980246514*ns)
    4.33  # result again returned in MeV
~~~~ 
passes.

eduardo-rodrigues:
That's right, @shantanu-gontia, thanks. There are quite a few of these to be changed, if one does go forward with doctests. This is probably a task for the medium term, once a few other bits are added to the package.

eduardo-rodrigues:
Just for my record, I should look at `pytest --doctest-modules --ignore-glob="*__main__.py"`.

eduardo-rodrigues:
Closed via https://github.com/scikit-hep/particle/pull/148.

PR #148, opened 2019-07-03
Examples update to pass doctests

eduardo-rodrigues:
Addresses most of https://github.com/scikit-hep/particle/issues/108. What remains to be done is an update of the CI so that doctests are run by default. @henryiii, what would be the best way?

For the record I ran doctests on all but the `test` directories and `__main__.py` with the command
`pytest --doctest-modules --ignore-glob="*__main__.py" --ignore-glob="test*"`.


eduardo-rodrigues:
Thanks @henryiii for the commit. All fine apart from the tests on Py 2.7, again because of Pandas not being installed. We can make it simple and only run doctests for Python 3? Seems enough.

henryiii:
We can do that for now.

eduardo-rodrigues:
OK, done that.

PR #150, opened 2019-07-30
Improve exception from, drop single six usage without requirements

### LINKED ISSUE #149: Brevity in error handling when unavailable pdgid value entered
A call to `from_pdgid` with an unavailable value, results in the following stack trace

```python
>>> from particle import Particle
>>> Particle.from_pdgid(600095)
Traceback (most recent call last):
  File "D:\Projects\particle\particle\particle\particle.py", line 497, in from_pdgid
    return table[table.index(value)]
ValueError: 600095 is not in list

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\Projects\particle\particle\particle\particle.py", line 499, in from_pdgid
    raise ParticleNotFound('Could not find PDGID {0}'.format(value))
particle.particle.particle.ParticleNotFound: Could not find PDGID 600095
```

Probably the error could be handled before the table lookup. The stack trace would be cleaner.

henryiii:
Fixes #149.

### DIFF
--- a/particle/particle/particle.py
+++ b/particle/particle/particle.py
@@ -493,16 +493,10 @@ C (charge parity) = {C:<6}  I (isospin)       = {self.I!s:<7}  G (G-parity)     
         if not is_valid(value):
             raise InvalidParticle("Input PDGID {0} is invalid!".format(value))
         table = cls.all()
-        try:
+        if value in table:
             return table[table.index(value)]
-        except ValueError: # from None (Python2 workaround below)
-            err = ParticleNotFound('Could not find PDGID {0}'.format(value))
-            # I think this could be six.raise_from(err, None)
-            # But we are not requiring six
-            if hasattr(err, '__supress_context__'):
-                err.__supress_context__ = True
-            raise err
-
+        else:
+            raise ParticleNotFound('Could not find PDGID {0}'.format(value))
 
     @classmethod
     def findall(cls, filter_fn=None, particle=None, **search_terms):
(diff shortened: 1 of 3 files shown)


henryiii:
Someone needs to verify the Python3 exception message is actually shortened.

particle/particle/particle.py
```diff
@@ -495,8 +495,13 @@ def from_pdgid(cls, value):
         table = cls.all()
         try:
```
henryiii: Could be `if value in table:` instead?

eduardo-rodrigues: I just tried and the exception message is not really shortened because it still gives the trace of `ValueError` and `ParticleNotFound`. I agree that a shorted trace can be achieve checking the existence of the value with `if value in table` rather than inside a `try` statement.

eduardo-rodrigues: I think we can merge this PR as soon as you make the change you suggest above. Seems appropriate not to have a "cascade" of 2 exceptions caught here - no gain.

henryiii: Is the new version better?

henryiii:
Also now requiring hepunits 1.0, this does not work with hepunits 0.1.1.

particle/particle/particle.py
eduardo-rodrigues: Yep, stack trace shorter hence clean(r)er.

Issue #135, opened 2019-05-15
Add a Particle.dump_table() method


eduardo-rodrigues:
This should provide a nice printout of the whole CSV data table loaded.

Optionally, one could add an argument to save the table.

eduardo-rodrigues:
Dealt with in https://github.com/scikit-hep/particle/pull/151.

PR #151, opened 2019-08-17
Addition of Particle.dump_table(...) method

eduardo-rodrigues:
Addresses https://github.com/scikit-hep/particle/issues/135.

### DIFF
--- a/particle/particle/particle.py
+++ b/particle/particle/particle.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import, division, print_function
 # Python standard library
 import operator
 import os
+import sys
 import re
 import csv
 from copy import copy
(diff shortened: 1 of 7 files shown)


eduardo-rodrigues:
Sorry @henryiii, I just noticed that I started this work from a local copy of what you have prepared for PR https://github.com/scikit-hep/particle/pull/150! You can still review this PR.

eduardo-rodrigues:
Note that `tabulate` has extra options and one could easily make those available via extra keyword arguments (`**kwargs`) to `dump_table` … Maybe for later.

eduardo-rodrigues:
For the sake of example, here is the kind of output one gets:

```
>>> Particle.dump_table(n_rows=5, exclusive_fields=['pdgid', 'pdg_name', 'mass'])
  pdgid  pdg_name      mass
-------  ----------  ------
      1  d             4.67
     -1  d             4.67
      2  u             2.16
     -2  u             2.16
      3  s            93
```

```
>>> Particle.dump_table(n_rows=5, exclusive_fields=['pdgid', 'pdg_name', 'mass'], tablefmt='rst')
=======  ==========  ======
  pdgid  pdg_name      mass
=======  ==========  ======
      1  d             4.67
     -1  d             4.67
      2  u             2.16
     -2  u             2.16
      3  s            93
=======  ==========  ======
```

particle/particle/particle.py
```diff
@@ -148,22 +148,22 @@ class Particle(object):
...
-    width = attr.ib()
-    anti_flag = attr.ib(converter=Inv)  # Info about particle name for anti-particles
-    _three_charge = attr.ib(Charge.u, converter=Charge)  # charge * 3
-    rank = attr.ib(0)  # Next line is Isospin
+    mass_upper = attr.ib(0.0)
+    mass_lower = attr.ib(0.0)
+    width = attr.ib(-1.)
+    width_upper = attr.ib(0.0)
+    width_lower = attr.ib(0.0)
+    Q_times_3 = attr.ib(Charge.u, converter=Charge)  # charge * 3
```
henryiii: Since we use charge instead of Q, shouldn't this be charge_times_3?

particle/particle/particle.py
```diff
@@ -203,6 +203,80 @@ def all(cls):
...
+        if n_rows >= 0:
+            tbl_all = tbl_all[:n_rows]
+
+        # Build all table rows
+        tbl = []
+        for p in tbl_all:
+            tbl.append([getattr(p, attr)for attr in tbl_names])
+
+        if filename:
+            with open(filename, 'w') as outfile:
```
henryiii: This should support both file-like objects and filenames.

henryiii:
I've rebased without my comment; you'll need to do `git fetch` followed by `git rebase --hard origin/eduardo-dump-table`, otherwise it will incorrectly try to merge.

eduardo-rodrigues:
I will deal with these 2 comments asap.

Otherwise, happy with the method name and the use of tabulate? I really like the printing capabilities and the fact that the package is rather small, so no heavy dependency introduced.

particle/particle/particle.py
eduardo-rodrigues: OK, adding a line for this.

particle/particle/particle.py
eduardo-rodrigues: I will revert back to `_three_charge` as we already have the `three_charge` property as part of the public API. My goal here was to produce a table with nice names, but that can be easily achieve another way.

henryiii:
Yes, tabulate looks safe to add, and I think the method name is okay.

eduardo-rodrigues:
OK, @henryiii, this is ready to go from my side. I will later add some trivial tests …

Fixes https://github.com/scikit-hep/particle/issues/135.

Issue #25, opened 2019-01-24
particle id translation


HDembinski:
The module [particletools](https://github.com/afedynitch/ParticleDataTool) from Anatoli Fedynitch has a strong overlap with `particle`. I hope he will join forces with us to develop this module further, merging the capabilities of `particletools` in `particle`.

One thing that `particletools` has is the ability to translate from/to PDGID to various other ID systems which are out there. For example, CORSIKA (the leading cosmic ray air shower simulation code in cosmic ray physics) has its own system, Pythia has its own system, EPOS (popular hadronic interaction model) has its own system, ...

It would be nice to add a `foreign_ids` system, which can translate from/to PDGID. The translation tables should be included in the module for convenience, and preferably in a way that makes it easy for users to provide additional tables.

Technically, you would need what is called a bimap in Boost, a map that allows fast lookup from both key and value. See:
https://stackoverflow.com/questions/1456373/two-way-reverse-map
http://paddy3118.blogspot.com/2009/04/bimap-bi-directional-mappingdictionary.html
There is currently no standard lib implementation of a bimap as far as I know. Using just a dict as suggested in some answers of SO, does not avoid collisions. Collision example: in system A gamma == 1, in system B u == 1. What should dict[1] return, gamma or u? The way out is to use different enum classes as keys. So in addition to PDGID, there could be PYTHIAID and EPOSID and CORSIKAID. Then the lookup with a normal dict would work, because the keys while having colliding numerical values, are now distinct types.

HDembinski:
So far, I am just brainstorming, but it looks to me that the most elegant way is to add further submodules under the `particle` directory. It currently has `pdgid`, then there could be `pythiaid` with a PYTHIAID class, `corsikaid` with a `CORSIKAID` class and so on and so forth.

HDembinski:
Btw, here is the doc of `particletools`, which explains its capabilities.
https://particledatatool.readthedocs.io/en/latest/index.html

eduardo-rodrigues:
@HDembinski, I like your ideas, as said before. We should get in touch with Anatoli asap …

I do have some urgent matters to sort out and won't be able to jump immediately with developments on this front, but would be super happy if you can start on your suggestions.

In general I do think we can make a stronger package profiting from particletools, as you say.

Thanks a lot!

eduardo-rodrigues:
[...]

PR #155, opened 2019-08-22
Introduction of PythiaID class and PDGID-PythiaID converter

eduardo-rodrigues:


eduardo-rodrigues:
This PR starts addressing https://github.com/scikit-hep/particle/issues/25 for Pythia.

In the end, we cannot go the suggested way of using a bi-directional dictionary for the ID maps. The reason is simple: that requires that all ID classes cannot be directly compared to integers, as otherwise `bimap[PDGID(1)] = PythiaID(1)` would be equivalent to `bimap[PythiaID(1)] = PDGID(1)` and the bimap would not be constructed; only one map would. But we do rely *a lot* on `PDGID(1)==1` being True. So a no-go. My implementation of the bi-directional map is not as elegant as the suggested `TwoWayDict`, but it is fast and meets our requirements, avoiding to break *lots* of existing code. Bottom line: ready for review in my opinion.

particle/converters/pythia.py
```diff
@@ -0,0 +1,48 @@
...
+    --------
+    >>> pyid = Pythia2PDGIDBiMap[PDGID(9010221)]
+    >>> pyid
+    <PythiaID: 10221>
+
+    >>> pdgid = Pythia2PDGIDBiMap[PythiaID(10221)]
+    >>> pdgid
+    <PDGID: 9010221>
+    """
+    def __init__(self):
```
henryiii: I would generalize this. Allow a user to pick two classes; they become A and B, and then they are selected in the `__getitem__`.

particle/converters/pythia.py
```diff
@@ -0,0 +1,48 @@
...
+        elif isinstance(value, PDGID):
+            return PythiaID(self._from_map[value])
+        raise MatchingIDNotFound("Matching PDGID-PythiaID for input {0} not found !".format(value))
+
+    def __len__(self):
+        """Returns the number of connections."""
+        return len(self._to_map)
+
+
+Pythia2PDGIDBiMap = __Pythia2PDGIDBiMap()
```
henryiii: Would become `Pythia2PDGIDBiMap = BiMap(PythiaID, PDGID)`

particle/pythia/pythiaid.py
```diff
@@ -0,0 +1,63 @@
...
+
+import csv
+
+from .. import data
+from ..pdgid import PDGID
+from ..exceptions import MatchingIDNotFound
+
+
+with data.open_text(data, 'pdgid_to_pythiaid.csv') as _f:
+    _bimap = {int(v['PYTHIAID']):int(v['PDGID']) for v in csv.DictReader(_f)}
```
henryiii: Shouldn't this use a converter?

particle/converters/pythia.py
henryiii: PS: I can make this change if you want me to in 1-2 days.

eduardo-rodrigues: Good idea. I will try and get to this this week, then, no worries ...

particle/converters/pythia.py
eduardo-rodrigues: Sure.

particle/pythia/pythiaid.py
eduardo-rodrigues: I did this here because of (1) speed, because the bimaps call the 2 ID class constructors for each item, and (2) tp prevent a circular dependency, since the BiMap class imports also the 2 ID classes.

particle/converters/pythia.py
eduardo-rodrigues: OK, @henryiii, I've just committed a generic bi-map class. Have a look. I can then trivially define `Pythia2PDGIDBiMap` as function of it.

eduardo-rodrigues:
Working on-off on this is not the best thing to do! Now I finally fixed things, including the doctests.

I will rewrite the PDGID-PythiaID converter based on `BiMap` once we agree on the latter …
I then have a PDGID-GeantID bimap in the pipeline … But let's first get this in ;-).

particle/converters/pythia.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) 2018-2019, Eduardo Rodrigues and Henry Schreiner.
+#
+# Distributed under the 3-clause BSD license, see accompanying file LICENSE
+# or https://github.com/scikit-hep/particle for details.
+
+from __future__ import absolute_import
```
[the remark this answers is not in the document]
eduardo-rodrigues: Indeed, see my example above ;-). As said, I wanted first that we agree on the generic solution. I will make the next/final move now that we both agree.

particle/converters/bimap.py
```diff
@@ -0,0 +1,82 @@
...
+        >>> bimap[PythiaID(10221)]
+        <PDGID: 9010221>
+
+        Advanced usage:
+        >>> # Either pass a file name or a file object
+        >>> from particle import data
+        >>> filename = data.open_text(data, 'pdgid_to_pythiaid.csv')
+        >>> bimap = BiMap(PDGID, PythiaID, filename)
+        """
+        self.__A = class_A
```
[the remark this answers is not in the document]
eduardo-rodrigues: Indeed not needed. I do it by default; maybe a habit from C++. No harm in being softer here, so will change, no problem.

eduardo-rodrigues:
Done with feedback.

PR #161, opened 2019-08-29
Introduction of PDGID - EvtGen name converter

eduardo-rodrigues:
Makes use of extension of converters `BiMap` class prepared in https://github.com/scikit-hep/particle/pull/159. Merge only on the latter is merged.


eduardo-rodrigues:
So the failures are, of course, because  this PR depends on https://github.com/scikit-hep/particle/pull/159. I thought that flagging it as "draft" would avoid triggering the CI :S. Never mind.

henryiii:
It's okay to have a draft PR fail. It's handy to have the CI sometimes. If you want to skip ci, add `[skip ci]` to the current commit's message. Be sure to remove it though when doing the final squash and merge!

eduardo-rodrigues:
Adding `[skip CI]` still does not work for me, unfortunately! At least I tried again a few days ago and the CI ran, see https://github.com/scikit-hep/particle/pull/153. No idea. A detail, albeit an annoying one.

eduardo-rodrigues:
Just for reference: yet another converter relevant to the discussion in https://github.com/scikit-hep/particle/issues/25.

PR #168, opened 2019-09-07
Fixes for PDG particle renames

eduardo-rodrigues:


eduardo-rodrigues:
Just noticed these changes when doing some more work on the converters and cross-checking bits and pieces in PDGLive.

PR #175, opened 2019-09-25
Add conversions master file

eduardo-rodrigues:
This is a "master conversions file" I have been compiling. It contains much of the information then split in the various CSV files we already have. The idea is that this can be used as the file from which the others can be produced and updated. It makes sure we have a single entry point for all particle MC IDs and necessary names - less prone to mistakes.

The code to produce the CSV files is yet to be added. I wanted to have a PR just for this important file.

Note that this master file in fact contains already more information than the one in the present CSV files. The latter will be updated asap.


eduardo-rodrigues:
BTW, I do realise that I still need to document all the new files I have committed for conversions. There is much to add in `particle/data/`.

henryiii:
Why is this a FWF? CSV (or space separated) is easier to read in various programs in general. I can use the python csv module, Excel, Pandas, JupiterLab preview, etc. with CSV files. The alignment can remain the same, but it would be nice to have the commas, I think.

eduardo-rodrigues:
I did it this way to have a file that's trivial to read by a human, which is a must for the kind of information it contains. But you are right, I can still use a CSV. Have a look at this new version; happy?

henryiii:
I like this better, since it now is easier to load into most software too, and still readable. :)

eduardo-rodrigues:
Yet another move forward dealing with the ideas in https://github.com/scikit-hep/particle/issues/25.

PR #176, opened 2019-09-25
Rename Particle.from_dec to Particle.from_evtgen_name and use converters

eduardo-rodrigues:
This greatly simplifies the good-old `Particle.from_dec(...)` method, now that we have converters in the package. It seems natural to rename the method to `Particle.from_evtgen_name` to follow the conventions of the other `_name` properties and methods.

### DIFF
--- a/tests/particle/test_decfilenames.py
+++ b/tests/particle/test_decfilenames.py
@@ -616,10 +616,13 @@ list_dec_but_not_in_pdt = [
  'Sigma_b0',
  'Upsilon_1(1D)',
  'Upsilon_1(2D)',
+ 'Upsilon_2(1D)',
  'Upsilon_2(2D)',
  'Upsilon_3(1D)',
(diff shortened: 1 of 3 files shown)


tests/particle/test_decfilenames.py
```diff
@@ -723,26 +729,7 @@
...
- "K'_10",
- "Xi'_b-",
- "Xi'_b0",
- "anti-K''*0",
- "anti-K'*0",
- "anti-K'_10",
- "anti-Xi'_b+",
- "anti-Xi'_b0"
-]
+list_undealt_with = []
```
henryiii: This could probably be removed now, right?

eduardo-rodrigues: In fact the tests may well need to be "refactored". But at least this can be removed, as you say. Will do asap.

PR #178, opened 2019-10-01
Add directional maps PDG2EvtGenNameMap and EvtGen2PDGNameMap

eduardo-rodrigues:
I had to add these directional maps because one cannot really make an easy bidirectional map for 2 classes that are strings, as I did for PDGID and PythiaID, for example. The code of all these mappings provides the same feeling. Eventually these maps will be used in the `DecayLanguage` package, having more maps between names than the particles in the PDG table, which is a must in order to deal with all particles found in decay files. In that respect the CSV file will soon be updated with those name mappings of particles not in the PDG table.


henryiii:
Why can't this read in as a PDGID, then produce the EvtGen output? We already have PDGName -> Particle (and EvtGen -> Particle), and I assume we can do Particle -> EvtGen?

eduardo-rodrigues:
There are a few things to keep in mind, which explain the need for these converters:

1. One can indeed do what you say, but that will not be efficient at all. If all you want is the corresponding name in PDG (EvtGen , so in decay file) given the EvtGen (PDG) name, why would one go via the match of the `Particle` instance, and then get the name?

2. The `Particle` class has a lot of functionality, as it should. But not all applications need a `Particle` class instance. Sometimes the converters are all one wants/needs. So worth having both.

3. Once this is in, the decay file parser will use these converters instead. It will make it faster :-).

4. I should add the very important fact that the particles available via our PDG table are only a subset of the particles that exist, and hence a subset of the particles that are relevant/defined in decay files. If `DecayLanguage` would use a search via the `Particle` class it would miss many particles, hence the code would not be able to provide the names we want in many cases. Using the converters directly that problem is "circumvented". (There is still the little issue of getting the actual particle HTML names for the dot graphs, but that can be sorted out since we have code to programmatically produce them … OK, some bits extra may be needed, but that's another story.)

5. One could worry about proliferation of converter files and that is why I put much effort in producing a "master conversions file", which is now committed. Every other file is to be produced from this master file, ensuring consistency by construction. Also, having a single point of entry for updates makes it easier for the maintainers.

Convinced :-)?

henryiii:
It doesn't take long to read in a master file, and you can build any structure you want in memory. You can use Particle to do it, in fact, then just save a mapping. Having it listed separately does not speed anything up, right? There should not be any other files besides the master file.

The particle file being converted to CSV is special, since it uses Pandas and we didn't want that to be a dependency. But it is not done this way for speed.

eduardo-rodrigues:
Right, @henryiii, I removed the new CSV file I had introduced with this PR. Things are now simpler, using the `particle/data/conversions.csv` file.

eduardo-rodrigues:
I hope you will have a bit of time to review this PR asap? This way I can go ahead and make a new major release, which is now due, given the fair amount of stuff that I added since 0.6.2. Thanks very much.

particle/converters/bimap.py
```diff
@@ -100,3 +103,99 @@ def __str__(self):
...
+
+        name_A = name_A.upper()
+        name_B = name_B.upper()
+
+        fieldnames = None
+        skipinitialspace = False
+
+        if filename is None:
+            filename = data.open_text(data, 'conversions.csv')
+            fieldnames = ['PDGID', 'PYTHIAID', 'GEANTID', 'PDGNAME', 'EVTGENNAME', 'LATEXNAME']
```
henryiii: This may be fine for now, but eventually it might be nice to generalize this so that it works with any file, and then these are specified when calling DirectionalMaps.

henryiii:
Do you need to disable setitem, so the mapping cannot be broken?

particle/converters/bimap.py
eduardo-rodrigues: Agreed. There is probably a bit more to streamline, now that I can make better use of the master conversions CSV file. But that will be for later, to release the fair amount of updates in the pipeline.

eduardo-rodrigues:
In fact no, see the following:

```
In [3]: EvtGen2PDGNameMap['J/psi'] = 'test'
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-10ead3181abe> in <module>
----> 1 EvtGen2PDGNameMap['J/psi'] = 'test'

TypeError: 'DirectionalMap' object does not support item assignment
```

PR #183, opened 2019-10-28
Funding acknowlegment

eduardo-rodrigues:
FYI @henryiii.


henryiii:
Very good! I need to be better at making sure this is present. It's there in boost-histogram, at least. :)

eduardo-rodrigues:
Yeah, same here. I've been "reviewing" bits and pieces in preparation for CHEP … ;-)

(Our packages Particle and DecayLanguage are easy for what concerns funding since we're the ones working on them. As others join in, we will have to enlarge the acknowledgements section, and that will be nice. See also the new page on the website.)

PR #188, opened 2019-11-18
Remove Python 3.4 and add 3.8 support

eduardo-rodrigues:
@henryiii, I removed Python 3.4 as I feel it's no issue for the next major release. I also rearranged a bit the `setup.py`, adding `python_requires`.


setup.py
```diff
@@ -64,7 +68,6 @@ def get_version():
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
```
henryiii: Do we not support Python 3.8?

eduardo-rodrigues:
Yeah, I was going to add 3.8 support in a separate PR, actually, but you got quicker and added it here. Let me then just edit the title.

eduardo-rodrigues:
Yeah, I was going to add 3.8 support in a separate PR, actually, but you got quicker and added it here. Let me then just edit the title.

Issue #190, opened 2019-11-19
Data files licenses?


tamasgal:
We'd like to create a Julia version of this wonderful package and wanted to ask if we could reuse your `data` files. With proper citation of course!

eduardo-rodrigues:
You are certainly welcome to do so. BTW, I will probably have minor updates to them, and probably need to improve a bit the doc on what they provide, so feel free to ask anything unclear.

Glad you like the package :-).

tamasgal:
Thanks for the prompt reply and permission! :)

What do you think about a version number somewhere in the header of the data files? We could then easily compare and also set up a CI on our side to check occasionally for updates.

henryiii:
I'd like to see a link to the Julia package when it's ready, I'm curious to see what it will look like. :)

tamasgal:
Sure! The project has been started ;)

eduardo-rodrigues:
> What do you think about a version number somewhere in the header of the data files? We could then easily compare and also set up a CI on our side to check occasionally for updates.

The CSV files read with `csv` have limitations so having a comment line would need special code for the `csv.dictReader`. As the files will (soon ;-)) be rarely updated I would hope that we can go withing versioning per se. This being said, you could have a check for every major new version?

tamasgal:
I can live with that, sure! I mean I can simply refer to the package version as a whole.

However, I also quickly created a pull request, so feel free to merge it if you think it's a good idea to have comments in the CSV files: https://github.com/scikit-hep/particle/pull/192

PR #192, opened 2019-11-19
Allow CSV files to have comments starting with #

tamasgal:
This is a straightforward patch to allow CSV to have comments.

A nicer approach which I can of course also quickly implement is to create a wrapper function for `csv.DictReader`.

The aim is to allow version information or other comments to be added to the CSV files.

### DIFF
--- a/particle/converters/bimap.py
+++ b/particle/converters/bimap.py
@@ -72,10 +72,12 @@ class BiMap(object):
 
         with filename as _f:
             self._to_map = {converters[1](v[name_B]):converters[0](v[name_A])
-                            for v in csv.DictReader(_f)}
+                            for v in csv.DictReader(l for l in _f
+                                                    if not l.startswith('#'))}
             _f.seek(0)
             self._from_map = {converters[0](v[name_A]):converters[1](v[name_B])
-                              for v in csv.DictReader(_f)}
+                              for v in csv.DictReader(l for l in _f
+                                                    if not l.startswith('#'))}
 
     def __getitem__(self, value):
         if isinstance(value, self.class_B):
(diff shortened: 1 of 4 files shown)


eduardo-rodrigues:
I was trying to avoid this but I guess it won't slow the reader a lot. What's your feeling about this @henryiii?

tamasgal:
Well, the amount of data getting parsed is very low and Python is actually really efficient with string parsing and comparisons.

Btw. I ran `py.test` a few hundred times on both the master and on the PR and there is no significant impact on the run time of the full test suite.

tamasgal:
Also if you care about performance I can improve that part with cache functionality if needed. A simple LRU cache decorated wrapper function for `csv.DictReader` would make sure that the parsing only happens once for every single file in the repository during runtime.

eduardo-rodrigues:
I were faster ;-). Thanks! I will get this in tomorrow, then, unless @henryiii has follow-ups. Then I should add some actual comments to the files … You may have noticed that I committed updates in the meantime, and more is to come ...

eduardo-rodrigues:
Did you not rebase instead? Seems you have brought along my latest commits … Anyway, back to work on the package tomorrow ...

tamasgal:
Sorry, I had too much parallel project opened... fixed ;) I do many physics stuff in GitLab and there we "squash commits", which is a basically a rebase on merge. Accidentally missed that here.

henryiii:
I'm working on materials for a mini-course on high performance Python tomorrow, and had just noticed that directly iterating in Python over a file (like is done here) does not impact performance (which is the main reason one might not want to do it this way). Since this is a generator, it doesn't increate the memory usage either. So it's fine with me.

eduardo-rodrigues:
Addresses https://github.com/scikit-hep/particle/issues/190.

PR #195, opened 2019-11-20
Add versioning to data CSV files

eduardo-rodrigues:
FYI @tamasgal and @henryiii, following from https://github.com/scikit-hep/particle/pull/192/.
How about this versioning "scheme"?

I profited from the occasion to rename a file so that it better follows the convention of the other CSV files.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -1,6 +1,18 @@
 Changelog
 =========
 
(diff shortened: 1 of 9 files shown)


tamasgal:
I like it!

eduardo-rodrigues:
Merging as there is source of controversy in this PR ;-).

Issue #194, opened 2019-11-19
Isospin for photons?


tamasgal:
The isospin `I` of photons is returning `<2`. It's bit confusing since `isospin` is related to strong interactions and W for example has no entry in the raw CSV file.

The `<2` seems to refer to a "weak isospin" which has an unfortunate naming and the same symbol `I`. Wondering how this could be cleaned up a little bit? For the Julia implementation we try to fix the types and the photon is the only exception.

I guess you also had a discussion before you decided to put a `<2` into the CSV `I` column? :wink: 

```python
>>> from particle import Particle
>>> g = Particle.from_pdgid(22)
>>> g.I
'<2'
```

eduardo-rodrigues:
Thanks for spotting this! The short answer is that it's for sure not meant to be returning `<2`. The longer answer (with explanations) will have to wait until I get back online with more time … later tonight, I guess ;-).

tamasgal:
OK, credits for spotting it goes to @8me btw. ;)

eduardo-rodrigues:
So, the longer explanation. As you've probably seen by now, the package provides a curated version of the PDG .mcd file, with a lot of functionality and goodies. As far as particle information is concerned, we provide extended info making use of an extended file the PDG released back in 2008, see `https://github.com/scikit-hep/particle/blob/master/particle/data/mass_width_2008.fwf`. Clearly, a lot has changed since then, from particle data to PDG IDs, etc., which explains why we've had to prepare an extension (with bug fixes) to the extended file, see `https://github.com/scikit-hep/particle/blob/master/particle/data/mass_width_2008_ext.fwf`. Nothing of this matters to the user, of course, given that the user "interacts" behind the scenes with a curated CSV file. 
I have no idea why the PDG set the photon isospin to <2. This I'm going to fix asap. I will also try and see if other particles need a hotfix ...

Beyond the hotfix: note that we have it on the to-do list to simplify things, as the extension to the extended file became rather large by now, see https://github.com/scikit-hep/particle/issues/118. As this is gymnastics it's been on a lower priority. I hope to get to this at some point in early 2020 at the latest …

As always, the more eyes and/or users one gets, the better the package 👍 ! Thanks to you and @8me.

tamasgal:
I see, thanks for the explanation! Starting to get a good overview about the structure, which of course helps a lot to provide a Julia package with similar functionality.
It's also already more than obvious that this is not just a simple CSV -> Python class thingy ;)

[...]

PR #196, opened 2019-11-20
Make the particle isospin property return a float

eduardo-rodrigues:
Fixes https://github.com/scikit-hep/particle/issues/194.


eduardo-rodrigues:
FYI @tamasgal I will prepare a new release as soon as this goes in, as it is important enough. This way you will have a better package to use. Do let us know if you need anything else, e.g. converters from PDG ID to other MC IDs that we do not yet include …

The release should happen tomorrow.

tamasgal:
Thanks for the info, we are currently totally fine with it as it is, but looking forward to the release!

eduardo-rodrigues:
@henryiii, OK for this important fix to be merged?

PR #203, opened 2019-11-26
Remove 2017 PDG data file mass_width_2017.mcd

eduardo-rodrigues:
There have been so many updates and fixes for obsolete/incorrect extended data in the PDG that I reckon it is best to remove this 2-year-old file. We can for now keep the 2018 one together with the latest file, 2019. I don't think it will harm anyone. Let me know if you think otherwise. Else this won't show up in the major release 0.8.0 that I will prepare this week.


eduardo-rodrigues:
Thoughts, @henryiii?

henryiii:
Until we hit 1.0, we probably don't need to keep old files. After that, we probably want to be able to load old files (unless the "backward-compatibility" we provide is just to load an older version of Particle).

eduardo-rodrigues:
OK. Agreed.

tamasgal:
I realised I need to dig deeper in your CI, so far we are building upon the CSV files 😉

eduardo-rodrigues:
I think you are safe, there, @tamasgal. The other files are their for reference and because they are the inputs we curate.You do not need to do the same curation.

Unless I misunderstood your comment, in which case you should drop a line ...

tamasgal:
No I think we are on the same line. So we stick to build upon your CSV then!

henryiii:
The CSV was always intended to be built upon; the original files are just for us, since we have to produce the CSVs from them anyway, and we have to update every year, we keep our conversion to CSV going and edit the original file or an extended file. Users were always supposed to see and build on the CSV's.

tamasgal:
Alright, thanks for the confirmation Henry!

PR #213, opened 2020-01-06
CSV files version 4

eduardo-rodrigues:
Various updates, additions, refixes to the data files.

@henryiii, it makes little sense to me to keep `particle2008.csv` in the repo given the huge amount of changes that were made from the PDG and our sides since then. I will remove this file and will merge tomorrow Tuesday unless you see a good reason not to - I don't.


eduardo-rodrigues:
Hmm, to keep PRs specific I will open an issue for the removal of `particle2008.csv` and won't deal with it here, as this PR is about an update to CSV files actually used internally by the package.

tamasgal:
Ah, this reminds me that I wanted to open an issue. Would you agree to also include a version comment into the particle files?

eduardo-rodrigues:
Hi @tamasgal, thanks for bringing this up! You're right, it makes very much sense given that we always end up in practice with several updates given a year PDG file.

BTW, most of these updates are for particles that are of no interest to you. They mostly concern badly-known/unconfirmed particles that nevertheless have to be smoothly dealt with when dealing with our EvtGen decay files and other MC programs. I hope I will be done with most updates soon :D.

eduardo-rodrigues:
I will work a bit more on the PR to (1) take your request into account and (2) add some particles that are needed for decay files.

eduardo-rodrigues:
@henryiii, let me ask you since you added `pre-commit`: how do I fix the failures? I guess there is a command I can run locally? Or can I also fix things directly here? Apologies if this is obvious, but it seems quicker to ask. Thanks.

eduardo-rodrigues:
> Perfect, thanks!

Welcome. (I just need to fix the one thing I always forget about for Python < 3.6 :D.)

henryiii:
The command locally is:

(`pip install pre-commit` if you haven't already)
```
pre-commit run --all-files
```

If you run `pre-commit install`, then it will happen automatically every time you commit, and will stop the commit if it doesn't pass, and will make the changes for you, so you don't have to think that much about it. [See more here](https://pre-commit.com).

The "remote" way of doing it was not considered "safe" yet for public repositories. see the note at https://github.com/pre-commit/action

eduardo-rodrigues:
Thanks @henryiii.

PR #220, opened 2020-02-11
Make Particle.dump_table() work better in notebooks

eduardo-rodrigues:
This way code such as
```
from IPython.display import HTML
HTML(Particle.dump_table(filter_fn=lambda p: p.pdgid.is_meson and p.pdgid.has_strange and p.ctau > 1 * meter, exclusive_fields=['pdgid', 'pdg_name', 'html_name'], tablefmt='html'))
```
works nicely.


henryiii:
:+1: It's almost always better to return a string than to print! One exception is if you need a terminal width for some reason.

PR #221, opened 2020-02-12
Deal with particles with no mass info

eduardo-rodrigues:
This is relevant to deal e.g. with diquarks.

While at it, the `Particle.empty()` dummy particle got fixed as well, since the definition was not completely what one would expect - it should provide no mass/width info.

### DIFF
--- a/particle/particle/particle.py
+++ b/particle/particle/particle.py
@@ -169,7 +169,7 @@ class Particle(object):
 
     pdgid = attr.ib(converter=PDGID)
     pdg_name = attr.ib()
-    mass = attr.ib()
+    mass = attr.ib(-1, converter=lambda v: None if v < 0 else v)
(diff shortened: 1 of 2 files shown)


eduardo-rodrigues:
@henryiii, have a look if you have 2 minutes, just to be sure. This is all pretty trivial and fair to me ;-). Thanks.

particle/particle/particle.py
```diff
@@ -686,13 +685,26 @@ def _width_or_lifetime(self):
                 width=str_with_unc(self.width, self.width_upper, self.width_lower)
             )
 
+    def _safe_mass(self):
```
henryiii: Could also be called `_str_mass`

henryiii:
Looks good to me!

particle/particle/particle.py
eduardo-rodrigues: It's a better name. Will change.

Issue #226, opened 2020-03-28
Deprecation warning in particle.py


HDembinski:
Hi, I am getting a deprecation warning in python3.6. I am using particle 0.9.2.
```
/usr/local/Caskroom/miniconda/base/envs/prompt_hadrons/lib/python3.6/site-packages/particle/particle/particle.py:65: DeprecationWarning: The usage of `cmp` is deprecated and will be removed on or after 2021-06-01.  Please use `eq` and `order` instead.
```

henryiii:
https://www.attrs.org/en/stable/changelog.html#id3

We are doing this manually anyway, but we can use order instead and require attrs 19.3+.

HDembinski:
I think you mean 19.2+ right?

henryiii:
Yes, I missed the 19.2 heading. 19.3 was very short...

HDembinski:
Fixed! :D

PR #227, opened 2020-03-28
fix attr.s deprecation

HDembinski:
Addresses #226

### DIFF
--- a/particle/particle/particle.py
+++ b/particle/particle/particle.py
@@ -62,7 +62,7 @@ def _isospin_converter(isospin):
 
 
 @total_ordering
-@attr.s(slots=True, cmp=False, repr=False)
+@attr.s(slots=True, eq=False, order=False, repr=False)
 class Particle(object):
     """
     The Particle object class. Hold a series of properties for a particle.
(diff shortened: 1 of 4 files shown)


henryiii:
You can also drop total_ordering and set these to true. Mostly though we need to set a min attrs version so we don't get broken installs.

HDembinski:
Thanks for looking into this, @henryiii . Anything I could improve?

henryiii:
I don't think so, looks good to me.

HDembinski:
Also here, should we wait for Eduardo to have a look or merge?

henryiii:
@eduardo-rodrigues I was pretty sure you'd like it, I mostly wanted to make sure you were tracking the major changes so you can write the changelog instead of me ;)

eduardo-rodrigues:
Fine, Henry :D.

I need to find a bit more time to finish my big updates for nuclei (some bits will require discussion among us) and then we can make a new major release ...

henryiii:
I'd probably like to make an update to the setup.cfg too before a new release.

Issue #229, opened 2020-03-28
Redundancy and inconsistency between setup.py and requirements.txt


HDembinski:
In setup.py, we have:
```
INSTALL_REQUIRES = [
    'enum34>=1.1; python_version<"3.4"',
    'importlib_resources>=1.0; python_version<"3.7"',
    "attrs>=17.4.0",
    "hepunits>=1.1.0",
]
```

In requirements.txt, we have:
```
enum34>=1.1; python_version<"3.4"
importlib_resources>=1.0; python_version<"3.7"
attrs
hepunits
tabulate
```

I propose to read the requirements in setup.py from requirements.txt to enforce consistency between them.

HDembinski:
I would also try to add a unit test to make sure that requirements-dev.txt is a strict superset of requirements.txt.

henryiii:
Why have a requirements.txt at all?

HDembinski:
I don't know :)

henryiii:
I'd add the dev requirements as a extras (`[dev]`) and remove these files, but maybe that's just me.

HDembinski:
https://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py

HDembinski:
Also relevant but not conclusive:
https://packaging.python.org/discussions/install-requires-vs-requirements/

HDembinski:
I like your solution, @henryiii

henryiii:
The point there in the comments is that install_requires has the minimum requirements to get the package to work. You should view requirements.txt as a package manager guideline - which is not necessarily the same thing, and doesn't often have a reason to sit in your package at all. If you are building a non-installable (no setup.py) package, like an analysis, then requirements.txt would be used to get and pin versions.

So I like getting rid of requirements.txt. We don't have to get rid of the dev one (as we are using it to set up a test environment), but extras_require is a nicer way to to that and can be accessed by users - requirements.txt are not accessible from the wheel 99% of users download via pip, but the extras are.

henryiii:
I can do this in the near future, along with moving the most of the code from setup.py to setup.cfg, if you want.

eduardo-rodrigues:
It's a good move. To be honest I had this in mind since a while but never got to the work - I should have made an issue a while back! Thanks for bringing this up.

HDembinski:
This can closed, I think?

PR #230, opened 2020-03-28
remove requirements.txt

HDembinski:
Following the discussion in #229 it seems best to remove requirements.txt and perhaps keep requirements-dev.txt for developers to set up the dev environment. An alternative was mentioned by @henryiii , to add a special target `[dev]` to setup.py for that purpose. I like to have a requirements-dev.txt, though, because this is a common way to setup the dev environment that many packages in Scikit-HEP and elsewhere already use. So following the principle of least surprise I would keep this file. requirements.txt is not necessary however, since pip can read the requirements from setup.py.

### DIFF
--- a/particle/particle/particle.py
+++ b/particle/particle/particle.py
@@ -62,7 +62,7 @@ def _isospin_converter(isospin):
 
 
 @total_ordering
-@attr.s(slots=True, cmp=False, repr=False)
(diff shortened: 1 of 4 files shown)


henryiii:
I know it's silly, but you'll need to rebase. Git considers this a conflict.

Issue #210, opened 2019-12-13
Nuclei are not handled


HDembinski:
```
(Pdb) p P.from_pdgid(1000130270)
*** particle.particle.particle.ParticleNotFound: Could not find PDGID 1000130270
```

HDembinski:
This is not a question, it was intended as a bug report.

eduardo-rodrigues:
I have indeed recently added PDG IDs for a bunch of nuclei - what I could find. The known values and conversions are at `https://github.com/scikit-hep/particle/blob/master/particle/data/conversions.csv`. But the particle information comes from the PDG data table, which we improve (already quite a bit), which explains why no nuclei data is available in the `Particle` class. It would not be that trivial to add all the info.

HDembinski:
I am using `particle` to get names for all the particles which occur in a standard pp LHCb simulation, apparently also nuclei are generated in some detector interaction.

eduardo-rodrigues:
In other words, it is not a bug. We already provide more than the PDG. Nothing is impossible but I would need help … My idea of having the conversions was precisely to circumvent the limitations of the PDG table. Often what is needed is only conversion of IDs for MC programs and names.

HDembinski:
If you don't have a table entry for nuclei, I would fall back to a standard solution. You can get the charge and mass approximately from the PDGID, right?

eduardo-rodrigues:
I have recently provided some maps that should help for what I mention below, see for example `EvtGenName2PDGIDBiMap`. Is this not enough for your usecase?

HDembinski:
As name you can set "Nucleus(27, 13)"

HDembinski:
No, because I think that `particle.from_pdgid` should not raise errors as much as it does. It should try harder to return something useful.

eduardo-rodrigues:
The PDGID does not encode anything related to mass. Charge it does, which you can get directly from the `PDGID(...)` ... `Particle` and `PDGID` are different beasts.

HDembinski:
It also encodes that there are 27 nucleons, from that you can compute the mass, approximately.

HDembinski:
You can also fill the charge = 13.

eduardo-rodrigues:
[...]

PR #231, opened 2020-03-29
Addition of information on nuclei

eduardo-rodrigues:
Fixes https://github.com/scikit-hep/particle/issues/210.

While adding information on nuclei, which adds *a lot*, I encountered a serious slow down of `Particle.load_table`. Turns out that comes from a bug in its logic, see the fix `if append and value in cls._table:`. Making the check only run when relevant makes the loading > 10X faster :-). Now adding nucleus info won't any longer be an issue.

I will be adding the actual info shortly, with minor updates here and there.

FYI @henryiii and @HDembinski.


HDembinski:
I am very happy to see this coming. I looked into the implementation and was surprised that there are only edits of one file. Is the DB already part of the repo?

eduardo-rodrigues:
Nope. I need to make some modifications to `Particle` and update the nuclei CSV file. I'm also checking various bits ... but I basically have a working version with me. We really are almost there ...

eduardo-rodrigues:
@HDembinski, I've just committed the file. I will need to add info on how it was created …

Would you be happy with this kind of output:
```
In [2]: print(Particle.from_pdgid(1000130270).describe())
Name: Al27           ID: 1000130270   Latex: $^{27}\mathrm{Al}$
Mass  = 25133.1437306 MeV
Width = ?
Q (charge)        = 13      J (total angular) = None     P (space parity) = ?
C (charge parity) = ?       I (isospin)       = None     G (G-parity)     = ?
    SpinType: SpinType.NonDefined
    Antiparticle name: Al27~ (antiparticle status: Barred)
```
?

eduardo-rodrigues:
For info here is how the CSV file of nucleus information got created:

- Use list of nuclei (and their isotopes) from the `periodictable` package.
- Use its masses and convert them to MeV scaling everything to account for a small difference in the mass of the neutron as provided by `periodictable` compared to what we have, as in the PDG data.
- Copy all information for the neutron and proton but with their corresponging "nucleus PDG ID":
  neutron: 2112 <-> 1000000010
  proton: 2212 <-> 1000010010
  which means that neutrons and protons appear twice, for completeness.

eduardo-rodrigues:
@henryiii, @HDembinski, I think this is now ready for discussion and review, modulo tests and minor things that will come later and separately.

eduardo-rodrigues:
Oh, @HDembinski - by the time nuclei are released in the package I hope you will consider to give us a star of appreciation :D.

eduardo-rodrigues:
As "background information": these updates to the PDG ID functions may at first seem strange, or an overkill, but they make things consistent given the special case of neutrons and protons in terms of PDG IDs. Indeed, atop the usual PDG IDs 2112 and 2212, also the "nucleus PDG IDs" 1000000010 and 1000010010 exist (see for example the `periodictable` package), and should be equivalent. This last update ensures this. (Admitedly it would have been easier if these special cases did not exist, but ...)

.pre-commit-config.yaml
```diff
@@ -7,7 +7,6 @@ repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v2.3.0
   hooks:
-  - id: check-added-large-files
```
henryiii: Why are we dropping this? You can [add a larger max size]
(https://github.com/pre-commit/pre-commit-hooks#check-added-large-files).

particle/particle/particle.py
```diff
@@ -378,8 +385,12 @@ def load_table(cls, filename=None, append=False):
             cls._table_names = []
 
         if filename is None:
-            filename = data.open_text(data, "particle2019.csv")
-            cls._table_names.append("particle2019.csv")
+            import fileinput
```
henryiii: This should be imported at the top - importing inside a function should only be done for special cases, a standard library module is just fine at the top.

particle/particle/particle.py
```diff
@@ -388,39 +399,46 @@ def load_table(cls, filename=None, append=False):
         else:
             cls._table_names.append("{0!r} {1}".format(filename, len(cls._table_names)))
 
-        with filename as f:
+        # The following 2 files are necessary for Python 2, otherwise simply
+        # with filename as f:
+        # works just fine in Python 3 !
+        from contextlib import closing
+
+        with closing(filename) as f:
```
henryiii: I assume closing on something that is already a context manager is fine?

.pre-commit-config.yaml
eduardo-rodrigues: > Let's get the updated build system in before we actually push 0.10.0, too.

Sure, no problem. The fact that I updated the CHANGELOG does not mean that I would push a release out. I remember you saying you have things in the pipeline. I just noticed the CHANGELOG was rather out-of-date.

eduardo-rodrigues: The file with all nuclei info is actually just over 500 KB, the limit. That line was not allowing me to commit :S. But it's most likely safe to re-add the line after this PR … (I had a quick look to see how to increase the limit to 1MB but did not immediately find a solution, hence went pragmatically as I just wanted to get the job done.)

particle/particle/particle.py
eduardo-rodrigues: OK, will get to this and the rest tomorrow. Let me call it a day - it's getting rather late.

Thanks for starting to take a look, BTW.

particle/particle/particle.py
eduardo-rodrigues: I found this workaround online and it seems it is precisely to solve thie kind of issue I got with Python 2 here.

.pre-commit-config.yaml
henryiii: `args: ['--maxkb=600']` (placed just below the id)

eduardo-rodrigues: Done, thanks.

particle/particle/particle.py
eduardo-rodrigues: Done putting the imports at the start of the file.

eduardo-rodrigues:
Both happy?

HDembinski:
Right, I gave you a star just now. :)

This looks all good. What happens when you do the old `Particle.find("p")`, do you get an error, because it is ambiguous now or does it return the 2212 proton? I think it should return the 2212 for backward compatibility.

HDembinski:
```
In [2]: print(Particle.from_pdgid(1000130270).describe())
Name: Al27           ID: 1000130270   Latex: $^{27}\mathrm{Al}$
Mass  = 25133.1437306 MeV
Width = ?
Q (charge)        = 13      J (total angular) = None     P (space parity) = ?
C (charge parity) = ?       I (isospin)       = None     G (G-parity)     = ?
    SpinType: SpinType.NonDefined
    Antiparticle name: Al27~ (antiparticle status: Barred)
```
This looks good to me. Is "?" a placeholder?

eduardo-rodrigues:
:-).

Yeah, so now it gives you
```
In [4]: Particle.findall(name="p")
Out[4]:
[<Particle: name="p", pdgid=2212, mass=938.272081 ± 0.000006 MeV>,
 <Particle: name="p", pdgid=1000010010, mass=938.272081 ± 0.000006 MeV>]
```
and `find` will tell you you have too many particles, since 2. This can be discussed but clearly one has 2 possibilities: either just have the proton (similar arguments for the neutron) just as 2212, or else have it identical but with PDG ID 1000010010, which is a valid ID used in various places for the H nucleu=proton, e.g. in `periodictable`. Personally I prefer to be more comprehensive, even if `Particle.findall(name="p")` now provides 2 replies.
I realise some documentation on nuclei needs to be included very soon ...

eduardo-rodrigues:
> In [2]: print(Particle.from_pdgid(1000130270).describe())
> Name: Al27           ID: 1000130270   Latex: $^{27}\mathrm{Al}$
> Mass  = 25133.1437306 MeV
> Width = ?
> Q (charge)        = 13      J (total angular) = None     P (space parity) = ?
> C (charge parity) = ?       I (isospin)       = None     G (G-parity)     = ?
>     SpinType: SpinType.NonDefined
>     Antiparticle name: Al27~ (antiparticle status: Barred)
> 
> This looks good to me. Is "?" a placeholder?

It is difficult to have consistency everywhere - super annoying - given that we have all kinds of missing or non-relevant quantum numbers depending on what we're talking about (the actual particles, the specific internal quantum numbers). In general the "?" means the information is not known/available.
In some cases you get a `None` to be consistent with that is returned by the ID functions, which mimic the best way possible what the C++ HepPID package was providing (for completeness let me say that `Particle` has a much more complete and less buggy version of the functions at this stage ;-)).

HDembinski:
Why not always return `None` when information is missing? This seems very Pythonic.

eduardo-rodrigues:
A concrete example to what I mean: think of G-parity: not relevant for most particles. One could go fancy and make a distinction between not available and non-relevant, but I suspect it will be quite difficult to deal with all use-cases … Certainly something to think about in the near future, as I agree some more consistency will be less confusing. Thanks for the ping on this.

eduardo-rodrigues:
> Why not always return None when information is missing? This seems very Pythonic.

So same as not relevant. @henryiii, any strong feelings? We can certainly try and make some changes, but then in a different PR ...

henryiii:
Without looking into it very closely, I'm in agreement with information missing == None.

eduardo-rodrigues:
Thanks both!

PR #235, opened 2020-04-01
New packaging system

henryiii:
Redesign the packaging system.

* Move `particle` to `/src/particle` - reduces the ugly workarounds for pytest, zipapp, and notebooks, including the symlink. Package *must* be installed to use (`python -m pip install -e .`) (you can still add the `src` dir to the PYTHONPATH, but as always not recommended)
* Use PEP 517/518 wheel builds
* Automatic versioning with `setuptools_scm` - no in-repository version file
* Better SDists, working MANIFEST.in and new check for contents with `check-manifest`
* Simpler setup.cfg, less custom code than setup.py
* Fixed importing files in tests (should come from data open_text, not hard paths)
* Uses extras, such as `pip install particle[convert]` or `pip install .[test]` to install extra optional requirements (no more requirements.txt-like files)
* Fix the Python 2 issue with the ZipApp (still broken, see #237).
* Move builds to GHA for sdist and ZipApp.


eduardo-rodrigues:
Hi Henry, this is technical and you're the expert here. No need to wait for a review from me; just merge once you're done and happy :-).

Nice to see better and more homogeneous things such as these throughout Scikit-HEP, BTW!

eduardo-rodrigues:
Concerning "Automatic versioning with setuptools_scm - no in-repository version file": this is the one improvement I cannot understand in Python: finding a version is (1) an important thing and (2) should be trivial. If the number disappears from a trivial file to look at, how can somebody browsing the files/repo find immediately the version?

henryiii:
> finding a version is (1) an important thing and (2) should be trivial.

If you have a git checkout, it is available, just ask git. If you build the package, it does exist - it's in src/particle/version.py. If you download the file from PyPI, in all forms, it's there, in src/particle/version.py.

The version now exists in one place (from git tags) so it can't be out of sync, and it respects *every* git commit; the current version is `0.9.3.dev14+g51f6710`. This tells me that it is 14 commits past version 0.9.3, and even tells the the correct git hash. This is both useful information that you could never track in an existing file, as well as important for development on some packages; if you tell pip to rebuild from a remote file, it caches on the version - if you didn't change 0.9.3 to something else, it will just reinstall the original 0.9.3 version! Chris Burr pointed me at the Python Packaging Authority's `setuptools_scm` tool specifically to solve this problem.

It also simplifies the release process - The goal is to have a single, GitHub "release" trigger everything needed, no points for the maintainer to mess up - this is working for the testdata package.

PS: For a manual search, I'd look in CHANGELOG anyway, and you can see it there.

henryiii:
PS: I'm okay to call the file _version.py instead of version.py; the only official access to the version is the common idiom `particle.__version__` or of course, the correct Python 3.8 way to get the version from an installed package, `from importlib.metadata import version; version("particle")` (`pkg_resouces` for older versions of Python). What do you think? I've already called it version.py in a couple of other packages, but we could change that.

eduardo-rodrigues:
Thanks for the info and clarifications.

Sure, go with what is recommended by the PyPA … Funny enough we had version.py before, and that then got changed to _version.py, IIRC. But go with the new standard.

eduardo-rodrigues:
Ready to be merged, then.

Issue #238, opened 2020-04-07
Laziness feature request: `from particle import literals`


HDembinski:
Literals are great and I want to use them a lot!

Currently, I need to import literals like this
`from particle.particle import literals as lp`
For the end user it is not obvious why one has to type the middle `particle`, so this is a bit confusing (I think I understand why you organised it like this, but just saying).

It would be a great convenience if I could just import my literals like this
`from particle import literals as lp`

tamasgal:
That would be a nice first PR 😉

henryiii:
Should it be something like `particle_literals`? There are also PDGID literals, which is why they are structured this way. Though I think particle should be the default/easy one.

eduardo-rodrigues:
Right, we could keep `from particle.pdgid import literals as lid` for `PDGID` literals and make the particle literals available from the top-level import? That's doable.

HDembinski:
That sounds like a good solution.

HDembinski:
Great, if I could I would give another star.

PR #239, opened 2020-04-15
Make particle literals available from top-level import

eduardo-rodrigues:
Fixes https://github.com/scikit-hep/particle/issues/238.

### DIFF
--- a/src/particle/__init__.py
+++ b/src/particle/__init__.py
@@ -21,7 +21,8 @@ from .particle import ParticleNotFound, InvalidParticle
 
 # Direct access to Particle literals
 from .particle import literals
-sys.modules['particle.literals'] = literals
+
+sys.modules["particle.literals"] = literals
 
 # Direct access to kinematics functions
 from .particle import width_to_lifetime, lifetime_to_width
--- a/src/particle/__init__.py
+++ b/src/particle/__init__.py
@@ -3,6 +3,8 @@
 # Distributed under the 3-clause BSD license, see accompanying file LICENSE
 # or https://github.com/scikit-hep/particle for details.
 
+import sys
+
 # Convenient access to the version number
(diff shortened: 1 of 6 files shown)


eduardo-rodrigues:
Now you can do
```
In [1]: from particle import literals as lp

In [2]: from particle.literals import Lambda_b_0
```

src/particle/__init__.py
```diff
@@ -17,6 +19,11 @@
 from .particle import Particle, SpinType, Parity, Charge, Inv, Status
 from .particle import ParticleNotFound, InvalidParticle
 
+# Direct access to Particle literals
+from .particle import literals
+
+sys.modules["particle.literals"] = literals
```
HDembinski, line 25: For my own education, why is this line needed?

eduardo-rodrigues:
> Looks great! I suppose you also test somewhere that from particle.pdgid import literals still works as before.

Correct, see `tests/pdgid/test_literals.py`.

src/particle/__init__.py
eduardo-rodrigues: This is needed for `from particle.literals import ...` to work if I do not move the submodule around. Indeed the statement expects `particle.literals` to be a file/module, and it isn't strictly speaking since at `particle.particle.literals` in reality. Indeed I thought I would keep the structure and old functionality. This line is needed at least to the best of my knowledge to make both statements work. Without it `from particle import literals as lp` would work but not `from particle.literals import Lambda_b_0`.

HDembinski: Ah, I see, so this is a pretty clever solution. Thanks for the nice explanation.

henryiii: :+1:

eduardo-rodrigues:
Thanks both!

PR #243, opened 2020-05-18
Preparing for release, add warnings

henryiii:
Getting closer to 0.10.


CONTRIBUTING.md
```diff
@@ -33,3 +33,9 @@ Use PyTest to run the unit checks:
 ```bash
 pytest
 ```
+
+# Developers: making a new release
+
+* Make sure the changelog is up to date.
+* Copy the contents of the changelog entry for that version
```
eduardo-rodrigues, line 40:
```suggestion
* Copy the contents of the changelog entry for that version.
```

CONTRIBUTING.md
```diff
@@ -33,3 +33,9 @@ Use PyTest to run the unit checks:
...
+* Make a new release on the GitHub Releases page, using the tag name `v#.#.#` (will be created)
```
eduardo-rodrigues, line 41:
```suggestion
* Make a new release on the GitHub Releases page, using the tag name `v#.#.#` (will be created).
```

henryiii:
Fixing a bunch of `ResourceWarning: unclosed file`'s, and getting a little closer to supporting typing. I'm manually testing context managers since that's a bit more readable for now than an unbroken multiple with - Python 3.9 is introducing a new parser (PEG) for the first time since Python 1, and 3.10 will finally start using features of that new parser, probably including finally allowing line breaks in a multi-with statement (since that's one of the reasons Guido finally caved in, along with fixing a bunch of hacks currently in place to support things not technically supported by the old LL(1) parser but expected by users).

Note: as a reminder, *all* opened files need to be in a context manager, or manually have `.close()` called even if an error is thrown (but no need to do both). Single line "shortcuts" generate warnings *and* are usually less readable than the two line version.

eduardo-rodrigues:
I take note of your good points. Thanks.

PR #244, opened 2020-05-20
(internal) Typing support

### LINKED ISSUE #237: ZipApp fixes
The ZipApp seems to be broken. We could add a test for it. Here's the current status:

Python 3:
```
 ~/g/s/particle   henryiii/dev  python3 ~/Downloads/particle.pyz search 123                                                                       150ms  Thu Apr  2 22:38:44 2020
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/henryschreiner/Downloads/particle.pyz/__main__.py", line 3, in <module>
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 208, in run_module
    return _run_code(code, {}, init_globals, run_name, mod_spec)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/henryschreiner/Downloads/particle.pyz/particle/__main__.py", line 52, in <module>
  File "/Users/henryschreiner/Downloads/particle.pyz/particle/particle/particle.py", line 801, in from_pdgid
  File "/Users/henryschreiner/Downloads/particle.pyz/particle/particle/particle.py", line 234, in all
  File "/Users/henryschreiner/Downloads/particle.pyz/particle/particle/particle.py", line 391, in load_table
AttributeError: '_io.BytesIO' object has no attribute 'name'
```

Python 2:
```
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/henryschreiner/Downloads/particle.pyz/__main__.py", line 3, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 182, in run_module
    mod_name, loader, code, fname = _get_module_details(mod_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 111, in _get_module_details
    __import__(mod_name)  # Do not catch exceptions initializing package
  File "/Users/henryschreiner/Downloads/particle.pyz/particle/__init__.py", line 13, in <module>
  File "/Users/henryschreiner/Downloads/particle.pyz/particle/geant/__init__.py", line 8, in <module>
  File "/Users/henryschreiner/Downloads/particle.pyz/particle/geant/geant3id.py", line 19, in <module>
  File "/Users/henryschreiner/Downloads/particle.pyz/particle/data/__init__.py", line 9, in <module>
  File "/Users/henryschreiner/Downloads/particle.pyz/importlib_resources/__init__.py", line 5, in <module>
  File "/Users/henryschreiner/Downloads/particle.pyz/importlib_resources/_compat.py", line 8, in <module>
ImportError: No module named pathlib2
```

I think this last issue is one that came up with newer versions of `importlib_resources` - we need to install pathlib2 now in the zipapp. The Python3 error seems to have come from a recent change where the datafile is being treated as an actual file - there is no promise in Python that files really will exist as a text file on a disk - when you are in a zip file, this is an example of a case where they are not real files.

### LINKED ISSUE #245: Import time
I've noticed that something that changed recently has caused the import time to skyrocket. (This could even just be in my branch). Running this in Python 3.7 with `python -X importtime` shows a whopping 10+ seconds devoted to `particle.particle.literals`!

henryiii:
Adding typing support. Using Python 2-compatible type comments.

Currently we just have the minimum + one file fully done (pidid/functions), and basic support on Particle and with some of the other functions.

 MyPy has caught a variety of potential crashes related to multiplying values with None, abs(None), etc. Whenever you return None as part of your API, that means it has to be checked at every step down the line...

We also now have full, type-checked support for using any object that `SupportsInt` (that is, has a `__int__()` special method), so you can now pass `Particle`s directly to any of these functions!

We need to rethink or replace our loops over functions - PDGID is the main suspect. There's several options: Maybe we can simply write `name = _functions.name` and that will work (might be a problem in Python 2). Maybe we can simply remove the duplication of functions and methods, and make sure they work as free functions too, that is, PDGID.name(x) at a replacement for PDGID(x).name ; that would only work on Python 3, however; 2 users would need to make the object to use the method. Finally, we could just duplicate this and say that this is mostly static, so having things in two places is not too bad (the computations would be in once place, just the wrappers would be duplicated).

This fixes the ZipApp (which now requires Python 3.5+, is a bit smaller, and has a simpler setup). Dropped usage of file input, which was really designed for changing files listed on the command line and didn't work with file objects. Added a somewhat internal "_name" parameter that will overwrite the recorded file name (since some file objects do not have names, like files in zip directories).

Fixes #245. Closes #237.

To be clear: This targets 0.10. External typing support (for dependent libraries) will go into a later version as what we have is a bit hacky, but it still fixed some possible bugs and helped in a speed up the code by a factor of 100, which isn't bad.


henryiii:
I think this is the last item before release.

.github/workflows/ci.yml
```diff
@@ -87,16 +98,36 @@ jobs:
...
     - name: Install requirements
-      run: python -m pip install enum34 pathlib2 importlib_resources attrs hepunits tabulate --target src
-
-    - name: Move enum to prevent name clash
-      run: mv src/enum src/enum34
+      run: python -m pip install attrs hepunits tabulate importlib_resources --target src
 
     - name: Make ZipApp
-      run: python -m zipapp -p "/usr/bin/env python" -o ../particle.pyz .
+      run: python -m zipapp -c -p "/usr/bin/env python3" -m "particle.__main__:main" -o ../particle.pyz .
```
eduardo-rodrigues, line 104: Just for my understanding - does this mean that the ZipApp will only be available for Python 3?

docs/CHANGELOG.md
```diff
@@ -27,7 +28,7 @@ May 18th, 2020
             time).
     - Converter script adapted to add to the produced data files
         particles not in the PDG data table, such as diquarks.
-- Redesigned packaging system.
+- Redesigned packaging system, GHA deployment
```
eduardo-rodrigues, line 31:
```suggestion
- Redesigned packaging system, GHA deployment.
```

docs/CHANGELOG.md
```diff
@@ -36,6 +37,8 @@ May 18th, 2020
     - Deprecation warning in `attr.s` fixed, requirement on minimal
         version of `attr` added.
     - Version tags now follow standard `v#.#.#` format.
+    - Some Python warnings fixed
```
eduardo-rodrigues, line 40:
```suggestion
    - Some Python warnings fixed.
```

docs/CHANGELOG.md
```diff
@@ -36,6 +37,8 @@ May 18th, 2020
...
+    - Some initial work on static type hints
```
eduardo-rodrigues, line 41:
```suggestion
    - Some initial work on static type hints.
```

src/__main__.py
```diff
@@ -1,3 +0,0 @@
-import runpy
```
eduardo-rodrigues, line 1: So I take it that this file is no longer needed? I trust you there - did not test locally.

src/particle/pdgid/functions.py
```diff
@@ -497,7 +528,11 @@ def three_charge(pdgid):
 
     if _extra_bits(pdgid) > 0:
         if is_nucleus(pdgid):  # ion
-            return 3 * Z(pdgid)
+            Z_pdgid = Z(pdgid)
+            if Z_pdgid is None:
```
eduardo-rodrigues, line 532: This in principle should never happen for a nucleus, by construction ;-). But the protection is harmless and avoids trouble if for some reason info was to miss for a given nucleus, so go with it 👍 .

docs/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ May 18th, 2020
 - `PDGID` class:
     - PDG ID functions extended to correctly and consistently deal
         with nuclei.
+    - Functions now accept any int-like, including Particle objects
```
eduardo-rodrigues, line 21:
```suggestion
    - Functions now accept any int-like, including Particle objects.
```

docs/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ Changelog
 Version 0.10.0
 --------------
 
-May 18th, 2020
+May 21th, 2020
 
 - `Particle` class:
     - Several improvements, in particular to better deal with nuclei
```
eduardo-rodrigues, line 10:
```suggestion
    - Several improvements, in particular to better deal with nuclei.
```

eduardo-rodrigues:
> I think this is the last item before release.

You did quite a lot for the new release in the end! Honestly, I would have done the type checking after the release … but now that's done :-).

eduardo-rodrigues:
> We also now have full, type-checked support for using any object that SupportsInt (that is, has a __int__() special method), so you can now pass Particles directly to any of these functions!

Oh, this then needs to be documented somewhere, probably in the README at least, with some example?
> We need to rethink or replace our loops over functions - PDGID is the main suspect. There's several options: Maybe we can simply write name = _functions.name and that will work (might be a problem in Python 2). Maybe we can simply remove the duplication of functions and methods, and make sure they work as free functions too, that is, PDGID.name(x) at a replacement for PDGID(x).name ; that would only work on Python 3, however; 2 users would need to make the object to use the method. Finally, we could just duplicate this and say that this is mostly static, so having things in two places is not too bad (the computations would be in once place, just the wrappers would be duplicated).

I'm not sure I understand this. For sure it makes sense to have the free functions and the identical methods in `PDGID`.

src/__main__.py
henryiii: It uses `-m "particle.__main__:main"` in the zip app command instead (which writes this file for you) and that way you can't accidentally put this into the PyPI package (which was happening after I moved the files because I forgot about it!)

.github/workflows/ci.yml
henryiii: Yes. It lets us drop hacks and workarounds, and drop some packages to make it smaller. Importlib resources was requiring some new dependency only in Python 2 making it harder. You don't need to use python3 for the rest of your work, you just need installed on your computer.

src/__main__.py
eduardo-rodrigues: OK. Subtle. Thanks.

.github/workflows/ci.yml
eduardo-rodrigues: OK!

src/particle/pdgid/functions.py
henryiii: You can also state "assert <thing that is always true>" to make MyPy happy. But each function must fill out all logical possibilities, which is nicer to read and verify for someone not looking at everything at once, so I think it's also better. It also does sometimes catch crashes before they happen!

eduardo-rodrigues: Indeed. Makes complete sense.

I have to learn how to use MyPy…!

henryiii: I will probably be writing a post over it so will ping you when I do that. :)

eduardo-rodrigues: I appreciate that, thanks!

henryiii:
Example:

    >>> from particle import PDGID
    >>>
    >>> pid = PDGID(211)
    >>> pid
    <PDGID: 211>
    >>> pid.is_meson
    True
    >>> pid = PDGID(99999999)
    >>> pid
    <PDGID: 99999999 (is_valid==False)>

For convenience, all properties of the ``PDGID`` class work as standalone functions in Python 3:

    >>> PDGID.is_meson(211)
    True

PR #248, opened 2020-07-14
Add PDG 2020 particle data table .mcd file

eduardo-rodrigues:
The 2020 .mcd file is now available from the PDG.
This trivial PR adds it to the repository *but* does not make it the default file, since I'm working on improving the CSV files. Soon we will be in a position to use the 2020 info by default ...


henryiii:
Should we trim trailing whitespace (you can do it with `pre-commit run --all-files`, or turn off pre-commit checking for the PDG files (by extension or location)?

eduardo-rodrigues:
Am trimming the spaces just now. I only see 2 places and it does not harm.

Issue #250, opened 2020-08-12
Bug report: both baryons and antibaryons have parity +1


redeboer:
Hello scikit-heppers,
Thanks for creating this nice package! It provides a useful interface to the PDG.

We are considering to use this package in the [`expertsystem`](https://pypi.org/project/expertsystem) package, but noticed a potential bug while trying to implement it (https://github.com/ComPWA/expertsystem/pull/190): it seems that all baryons and antibaryons have parity +1.

For example, in the following, both have `P (space parity) = +`:

```bash
python3 -m particle search 2212  # proton
python3 -m particle search -2212  # anti-proton
```

For a list of anti-baryons that (I guess) should have parity -1, see:

```python
from particle import Particle

Particle.findall(lambda p: p.pdgid.is_baryon and p.pdgid < 0 and p.P == +1)
```

This seems to be the case in both the [latest release](https://github.com/scikit-hep/particle/releases/tag/v0.10.0) and the latest commit cae76410d6c45e8c2df4982ae6edb6735c544b41.

eduardo-rodrigues:
Thank you for reporting! Good catch! I really wonder how we never realised that. It must be that we have no test for it and the extended file that provides things such as parity (not available in the standard .mcd file from the PDG) is not being digested properly for baryons. I will look into this asap ...

eduardo-rodrigues:
BTW @redeboer, you are more than welcome to contribute to this package to augment what is provided by `Particle`and/or `PDGID`. I have myself a list of small bits of functionality that could be added at some point but maybe your package has extra needs? We can discuss this in a separate "issue" if you like.

redeboer:
Thanks, sounds good!
Internally, our package is just about propagating quantum numbers through a decay topology. Eventually, that decay topology is converted into an amplitude model for Partial Wave Analysis. It's only at the 'outer edges' of that topology -- the initial and final state -- that more general particle info like particle names comes into play (for the user interface). Practically speaking: we only require (conservation of) quantum numbers, which may clash with the concept of the `particle.Particle` class, because that contains more general information, like anti-particle and quark content.

But for sure I'll go through your source code if there I run into other issues, or post issues if I have some ideas!

redeboer:
[...]

PR #252, opened 2020-08-13
PDG particle data table 2020 update and made default

eduardo-rodrigues:


eduardo-rodrigues:
Have a look @henryiii if you have a bit of time today. Otherwise I will go ahead with the release for the ComPWA colleagues, see https://github.com/scikit-hep/particle/issues/250.

eduardo-rodrigues:
Thanks @henryiii!

PR #258, opened 2020-08-19
PDG ID qualification functions - enhancements, extensions, fixes for special particles, more doc

eduardo-rodrigues:


src/particle/pdgid/functions.py
```diff
@@ -165,9 +168,10 @@ def is_baryon(pdgid):
         return True
     if _extra_bits(pdgid) > 0:
         return False
-    if _fundamental_id(pdgid) > 0 and _fundamental_id(pdgid) <= 100:
+    if _fundamental_id(pdgid) in range(1, 101):
```
henryiii, line 171: IIRC, in Python 3, this is optimized to basically do the check above. :) (In Python 2, it generates the 100 numbers and then checks each on, since range is not a generator)

src/particle/pdgid/functions.py
```diff
@@ -369,7 +375,7 @@ def is_dyon(pdgid):
         return False
     if _digit(pdgid, Location.Nr) != 1:
         return False
-    if _digit(pdgid, Location.Nl) != 1 and _digit(pdgid, Location.Nl) != 2:
+    if _digit(pdgid, Location.Nl) not in (1, 2):
```
henryiii, line 378:
```suggestion
    if _digit(pdgid, Location.Nl) not in {1, 2}:
```

Set checking is faster (okay, maybe not even measurably faster when there are only two items, but on principle it's better). You could also use range.

src/particle/pdgid/functions.py
eduardo-rodrigues: Oh, then I should revert back. Good point.

henryiii: I think it's okay to not worry about optimizing for Python 2. But let me make a suggestion (will in a minute)

eduardo-rodrigues: OK, I will wait for your green light / suggestion. Otherwise I'm done here.

src/particle/pdgid/functions.py
```diff
@@ -163,21 +166,29 @@ def is_baryon(pdgid):
     # needs to be checked first since _extra_bits(pdgid) > 0 for nuclei
     if abs(int(pdgid)) in (1000000010, 1000010010):
         return True
+
     if _extra_bits(pdgid) > 0:
         return False
-    if _fundamental_id(pdgid) > 0 and _fundamental_id(pdgid) <= 100:
+
+    fid = _fundamental_id(pdgid)
+    if fid > 0 and fid <= 100:
```
henryiii, line 174:
```suggestion
    if 0 < fundamental_id(pdgid) <= 100:
```

src/particle/pdgid/functions.py
```diff
@@ -163,21 +166,29 @@ def is_baryon(pdgid):
...
         return False
-    if abspid(pdgid) == 2110 or abspid(pdgid) == 2210:
+
+    # Old codes for diffractive p and n (MC usage)
+    if abspid(pdgid) in (2110, 2210):
```
henryiii, line 177:
```suggestion
    if abspid(pdgid) in {2110, 2210}:
```

src/particle/pdgid/functions.py
```diff
@@ -267,11 +278,13 @@ def is_sm_gauge_boson_or_higgs(pdgid):
...
 
-    Codes 21-30 are reserved for the Standard Model gauge bosons and the Higgs.
+    Codes 21-30 are reserved for the Standard Model gauge bosons and the Higgs,
+    but only the codes 21-25 actually correspond to SM particles.
     """
     if abspid(pdgid) == 24:  # W is the only SM gauge boson not its antiparticle
         return True
-    return True if pdgid in range(21, 31) else False
+
+    return True if pdgid in range(21, 26) else False
```
henryiii, line 286:
```suggestion
    return True if 22 <= pdgid <=  26 else False
```

(You can use any combination of < and <=, depends on what is clearest)

henryiii: Keeping the range is perfectly fine too (here and below), on Python 3 it's just as fast.

src/particle/pdgid/functions.py
```diff
@@ -695,26 +721,27 @@ def S(pdgid):
...
+
+    if not (js == 1 or js >= 3):
         return 0
-    elif nl == 1 and js >= 3:
+
+    if nl == 0:
+        return 0 if js == 1 else 1
+    elif nl == 1:
+        return 1 if js == 1 else 0
+    elif nl in (2, 3):
```
henryiii, line 740:
```suggestion
    elif nl in {2, 3}:
```

src/particle/pdgid/functions.py
eduardo-rodrigues: Happy to accept this same suggestion but 21 should be included ;-).

eduardo-rodrigues: Should be 21-25.

eduardo-rodrigues: I will accept the suggestion and will correct it. Then will push corrections for the CI.

eduardo-rodrigues:
I'm done with the updates @henryiii. There is an issue with MyPy given your updates, though, and I do not know how to act. Any idea?

eduardo-rodrigues:
OK @henryiii, I pushed some more updates exploiting your suggestions. Done now :-). Thanks again for the help.

src/particle/pdgid/functions.py
```diff
@@ -129,13 +129,16 @@ def is_meson(pdgid):
         return False
     if abspid(pdgid) <= 100:
         return False
-    if _fundamental_id(pdgid) <= 100 and _fundamental_id(pdgid) > 0:
+    if 0 < _fundamental_id(pdgid) <= 100:  # type: ignore
```
henryiii, line 132:
```suggestion
    if 0 < int(_fundamental_id(pdgid)) <= 100:
```

This will make mypy happy, and not drop type checking for these lines. SupportsInt doesn't require that comparisons be defined, which is why this is complaining.

henryiii: Technically, this is a valid input to any of these functions:

```python
class MyVal:
    def __int__(self):
        return 3
```

But these functions with the comparisons will break if you don't call int() first, which is why mypy is (correctly) complaining.

henryiii:
What about the other type ignores? I can fix them if you prefer. After that, it's good to go.

eduardo-rodrigues:
I will do it shortly, will push and then merge. Thanks.

Issue #233, opened 2020-03-30
Design issue: tight tool coupling


HDembinski:
I looked into the `Particle.dump_table` method and saw some worrisome design choices, and so I wanted to write this down. To make it brief, adding convenience features is tempting, but it leads to interface and implementation bloat and one should really think twice about it. As a general rule, instead of calling some external tool inside your own interface, one should instead prepare the data in ones own tool so that the other external tool can easily consume it. This leads to a decoupling between Python packages, more combinatorical options for the user, and reduces interface, implementation, and unittest bloat.

This is a bit abstract, so let me point out what I mean specifically.

`dump_table` calles `tabulate` internally to make nice tables of the Particle data. This is nice, yes, but the same effect could have been obtained by just preparing the table data as a dict in Particle, and let the user pass this data to `tabulate`
```
tabulate(Particle.dump_table2(<options not related to tabulate>), <tabulate options>)
```
It is a really good idea to use `tabulate`, and this package can be recommended in the documentation, shown in examples, tutorials, mentioned in the FAQ. The call is an oneliner, there is barely more to type here than before.

But as the maintainer of Particle you save yourself a lot of work:
*You don't have to add, forward, document, and unittest keywords that change the output of `tabulate`, because the user is calling `tabulate` now and they can look up the options for `tabulate` themselves.*

If there is ever a package `super_tabulate` that is suddenly all hip with the kids, you don't have to do anything, users can just use that instead of `tabulate` (assuming that `super_tabulate` accepts the same input).

For me one of the key ideas of Scikit-HEP as a modular collection of specialized tools is to reduce direct coupling of tools like in `dump_table`, but offer generic and clear interfaces to pass data from these tools. This gives you enormous combinatorial power, and it reduces the implementation, documentation, and unittest load of each individual tool.

The fact that `dump_table` also supports a `filename` keyword to write to a file is just another variant of the same problem. Writing strings to a file is really easy in Python, you don't have to offer a functionality for that. Yes, it allows to quickly dump the output to a file, but it does not scale. What if I want to write it into a pipe instead of a normal file? What if I want to append the table to a file that has already some content? There are many ways to customize the action "write to a file-like object". If you try to support all those options, you need to add so many options to your interface. Or you just don't offer this shortcut. In Python there should be one way to do it. Writing stuff to files means you do `with open(filename, "w") as f: f.write(stuff)`. It is simple for simple things and scales to really complex things. And in Particle you only have to provide `stuff`.

[...]

PR #261, opened 2020-08-21
Replace Particle.dump_table with Particle.to_dict to remove strong coupling in that method

eduardo-rodrigues:
Addresses https://github.com/scikit-hep/particle/issues/233.

### DIFF
--- a/src/particle/particle/convert.py
+++ b/src/particle/particle/convert.py
@@ -110,7 +110,8 @@ def filter_file(fileobject):
             stream.write(line)
     stream.seek(0)
 
-    fileobject.close()
+    if not fileobject.closed:
+        fileobject.close()
 
     return stream
 
--- a/src/particle/particle/convert.py
+++ b/src/particle/particle/convert.py
@@ -161,6 +161,8 @@ def get_from_pdg_extended(filename, latexes=None):
         comment="#",
     )
 
+    filename.close()
+
     # Read the LaTeX
     latex_series = pd.concat([get_from_latex(latex) for latex in latexes])
 
@@ -294,6 +296,8 @@ def get_from_pdg_mcd(filename):
             "NameCharge",
         ),
     )
+
+    filename.close()
 
     ds = []
     for i in range(4):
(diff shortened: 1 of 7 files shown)


eduardo-rodrigues:
I will merge this tomorrow unless you have any feedback.

src/particle/particle/convert.py
```diff
@@ -109,6 +109,9 @@ def filter_file(fileobject):
         if not line.lstrip("\ufeff").lstrip().startswith("*"):
             stream.write(line)
     stream.seek(0)
+
+    fileobject.close()
```
henryiii, line 113: Shouldn't you only close it if you opened it?

src/particle/particle/particle.py
```diff
@@ -316,53 +312,68 @@ def dump_table(
...
+
+        >>> query_as_list = Particle.to_list(exclusive_fields=['pdgid', 'pdg_name'], n_rows=5)
+        >>> from tabulate import tabulate    # doctest: +SKIP
+        >>> print(tabulate(query_as_list, headers='firstrow'))    # doctest: +SKIP
+
+        Request the properties of a specific list of particles:
+
+        >>> query_as_list = Particle.to_list(filter_fn=lambda p: p.pdgid.is_lepton and p.charge!=0, exclusive_fields=['pdgid', 'name', 'mass', 'charge'])
+
+        >>> print(tabulate(query_as_list, headers='firstrow', tablefmt="rst", floatfmt=".12g", numalign="decimal"))    # doctest: +SKIP
```
henryiii, line 358: You can add tabulate as a doc requirement, then not skip this.

src/particle/particle/convert.py
eduardo-rodrigues: Absolutely. It is a possibility to call the method with a file opened. I will add a check.

src/particle/particle/particle.py
eduardo-rodrigues: OK, done.

Issue #259, opened 2020-08-20
Discussion on new PDGID qualifiers such as is_quarkonium


eduardo-rodrigues:
Hi @HDembinski and @henryiii. I'm working towards a version 1.0 of the package, dealing with outstanding issues and committing various things I've had in the pipeline.

While producing for inspection a series of tables per particle "category" I realised that it might be neat to make the several
"helpers" (I had to write) readily available to the user - OK, some are trivial, but not all. So I would enhance the `PDGID` class with several boolean functions beyond what we already have, such as `is_lepton` or `is_diquark`. This is the kind of list I'm thinking about:

```
- is_sm_quark
- is_sm_lepton
- is_gauge_boson_or_higgs
- is_special_particle
- is_heavy_flavor
- is_light_meson = is_meson and not is_heavy_flavor
- is_strange_meson
- is_charmed_meson
- is_charmed_strange_meson
- is_bottom_meson
- is_bottom_strange_meson
- is_bottom_charmed_meson
- is_quarkonium or is_heavy_qqbar_meson
- is_light_baryon
- is_strange_baryon
- is_charmed_baryon
- is_bottom_baryon
```

It's basically giving the class a means to straightforwardly provide the classification the PDG uses in their MC numbering scheme document  https://pdg.lbl.gov/2020/reviews/rpp2020-rev-monte-carlo-numbering.pdf. This can be used as usual in searches with `Particle.findall(...)`. Adding this ensures the user does not have to implement it itself and may also prevent silly bugs ...

What do you think? Thanks.

eduardo-rodrigues:
For the sake of example here is the kind of table I produced:

![table](https://user-images.githubusercontent.com/5013581/90756758-0f443700-e2dd-11ea-85d8-0ad3cbdc4fbb.png)

HDembinski:
I am on the fence on this. I can see how these are useful, but if you add these, you are adding a lot of redundant interface. Users should be empowered to compose these filters by themselves. The consensus in books on design and programming is that redundancy should generally be avoided in programming and in interfaces, because it easily gets out of hand and it adds "noise" to an otherwise clear and concise interface. More interface requires more documentation and testing. Plus, it can make it overwhelming to learn a tool which has a lot of interface that is actually redundant (you won't realize this on first glance).

I would rather spend a section in the docs to teach users how to compose these queries instead of adding pre-defined filters.

eduardo-rodrigues:
[...]

PR #262, opened 2020-08-22
More PDG ID qualification functions

eduardo-rodrigues:
Fixes https://github.com/scikit-hep/particle/issues/259.

### DIFF
--- a/src/particle/pdgid/__init__.py
+++ b/src/particle/pdgid/__init__.py
@@ -73,7 +73,6 @@ from .functions import is_lepton
 from .functions import is_hadron
 from .functions import is_meson
 from .functions import is_baryon
-from .functions import is_heavy_flavor
 from .functions import is_pentaquark
 from .functions import is_gauge_boson_or_higgs
 from .functions import is_sm_gauge_boson_or_higgs
--- a/src/particle/pdgid/__init__.py
+++ b/src/particle/pdgid/__init__.py
@@ -68,13 +68,17 @@ from __future__ import absolute_import
 from .functions import is_valid
 from .functions import abspid
 
+from .functions import is_quark
 from .functions import is_lepton
 from .functions import is_hadron
 from .functions import is_meson
 from .functions import is_baryon
+from .functions import is_heavy_flavor
 from .functions import is_pentaquark
+from .functions import is_gauge_boson_or_higgs
 from .functions import is_sm_gauge_boson_or_higgs
 from .functions import is_generator_specific
+from .functions import is_special_particle
 from .functions import is_nucleus
 from .functions import is_diquark
 from .functions import is_Rhadron
--- a/src/particle/pdgid/functions.py
+++ b/src/particle/pdgid/functions.py
@@ -20,7 +20,8 @@ This module provides the following:
 References
 ----------
 
-HepPDT and HepPID versions 3.04.01: http://lcgapp.cern.ch/project/simu/HepPDT/
+- PDG document "Monte Carlo Particle Numbering Scheme".
+- HepPDT and HepPID versions 3.04.01: http://lcgapp.cern.ch/project/simu/HepPDT/
 """
 
 from __future__ import print_function, division, absolute_import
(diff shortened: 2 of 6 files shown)


src/particle/pdgid/__init__.py
```diff
@@ -68,13 +68,17 @@
 from .functions import is_valid
 from .functions import abspid
 
+from .functions import is_quark
 from .functions import is_lepton
 from .functions import is_hadron
 from .functions import is_meson
 from .functions import is_baryon
+from .functions import is_heavy_flavor
```
HDembinski, line 76: Is "heavy flavor" really unambiguous? Does it include only c and b or also s? And what about top? If it is not completely obvious, then I would not expose this. Better  teach people to select based on the quark content.

src/particle/pdgid/__init__.py
```diff
@@ -68,13 +68,17 @@
...
 from .functions import is_pentaquark
+from .functions import is_gauge_boson_or_higgs
 from .functions import is_sm_gauge_boson_or_higgs
 from .functions import is_generator_specific
+from .functions import is_special_particle
```
HDembinski, line 81: Is the difference between special particle and generator specific clear?

src/particle/pdgid/functions.py
```diff
@@ -210,6 +221,16 @@ def is_diquark(pdgid):
     return False
 
 
+def is_heavy_flavor(pdgid):
```
HDembinski, line 224: See my comment before. The term "heavy flavor" is explicitly defined here, that is good, but I would still avoid it and rather teach people to use `has_charm` etc explicitly.

src/particle/pdgid/__init__.py
eduardo-rodrigues: Yes it is unambiguous - c,b,t. If we wanted to be totally strict I would say that nobody is 100% consistent in the sense that one should also include fourth-generation q-qbar states, if they were to exist. This 4th generation is encoded in the PDG scheme but not considered in hadron when saying heavy flavour hadrons, etc.

So yes, one can do without this definition, though the definition from the PDG doc is clear about what it means (and I made sure the docs provide the necessary information to the best of my knowledge).

src/particle/pdgid/__init__.py
eduardo-rodrigues: I simply took the classification from the PDG MC numbering scheme document. To be fair, I do not expect most people to ever have to care about these, and generator-specific particles/concepts are the most relevant ones. But, since the definition is not trivial - special set of particles - I thought I should provide it. In the doc I give examples and your comment makes me think that I should rather give the full list of "special particles" taken into account. I will update the doc a bit.

src/particle/pdgid/functions.py
eduardo-rodrigues: This def is certainly the one that is less important to have. I can leave it out, sure.

In the end, let's be fair, we will see how users react and it will always be possible to add functionality. Though I take your point that addings lots of extra is_XXX is not going to help having a clear and clean API.

eduardo-rodrigues:
@henryiii, happy with @HDembinski's suggestion? If yes I will remove that one method as I have no strong feeling.

src/particle/pdgid/functions.py
henryiii: I do see this as potentially rather handy, but I agree, let's keep the API simple at first, and see if this gets requested. The simpler building blocks are explicit and clear, and reduce the API that a user has to know to use it.

eduardo-rodrigues: OK, let's go with that. Thanks.

src/particle/pdgid/__init__.py
eduardo-rodrigues: I further developed the docstring. It should be very clear now what is meant.

eduardo-rodrigues:
Done implementing your suggestions, @HDembinski and @henryiii. Can at least one of you approve unless you have further feedback?

HDembinski:
I approve

PR #273, opened 2020-11-06
Minor simplifications in some PID functions

eduardo-rodrigues:


eduardo-rodrigues:
OK, finally ready for review.

src/particle/pdgid/functions.py
```diff
@@ -52,12 +52,14 @@ class Location(IntEnum):
 def is_valid(pdgid):
     # type: (PDGID_TYPE) -> bool
     """Is it a valid PDG ID?"""
-    if _fundamental_id(pdgid) > 0:
+    if _fundamental_id(pdgid) > 0:  # function always returns a positive number
```
henryiii, line 55: This comment is slightly confusing. I take it that it returns 0 or a positive number? If so, would `if _fundamental_id(pdgid) != 0:` or even `if _fundamental_id(pdgid):` be clearer?

henryiii (approved):

src/particle/pdgid/functions.py
eduardo-rodrigues: Correct, its >= 0, in other terms the abs of the PDG ID. I will amend to avoid the confusion.

PR #274, opened 2020-11-09
Add test for Particle.is_unflavoured_meson

eduardo-rodrigues:


src/particle/particle/particle.py
```diff
@@ -749,17 +749,19 @@ def is_unflavoured_meson(self):
         """
         pid = self.pdgid
 
-        if not pid.is_meson:
+        if not pid.is_meson:  # type: ignore
```
henryiii, line 752: Why are you just silencing mypy? What's the problem it reports, can't it be properly fixed instead?

eduardo-rodrigues: It's a fair question, @henryiii. I'm not yet mypy literate, unfortunately, hence only followed the other examples of such a statement in the same file. I would be happy to know if there's an easy way to not have mypy ignore such statements.

eduardo-rodrigues: The messages from mypy were
```
src/particle/particle/particle.py:752: error: "PDGID" has no attribute "is_meson"
src/particle/particle/particle.py:755: error: "PDGID" has no attribute "has_charm"
src/particle/particle/particle.py:755: error: "PDGID" has no attribute "has_bottom"
src/particle/particle/particle.py:755: error: "PDGID" has no attribute "has_top"
src/particle/particle/particle.py:762: error: "PDGID" has no attribute "has_strange"
```
and I guess this comes from the fact that the `PDGID` class gets such properties decorated. Still, no idea how to make mypy understand. Did you come across such things in the past?

henryiii: The problem is we add them dynamically, which is ugly especially for type checkers. It also means tools like PyCharm will not auto-complete them for you. I didn't realize that was the problem - we can skip for now, but in the future we could add a .pyi file with the static class structure in it, since it's doesn't change rapidly. We don't need this workaround in Python 3 - class methods are usable by themselves, so there's no need to duplicate.

eduardo-rodrigues: OK, then we proceed.

BTW, I would not call the procedure ugly. I did that dynamic addition to add with 3 lines of code a whole bunch of is_XXX and has_XXX functions defined as well as standalone functions.

henryiii: It's "ugly" because it loses valuable static information. Just like `from numpy import *` is ugly. Yes, it does a lot in very little code, but someone trying to figure out where some function comes from will struggle, because it no longer is self descriptive. I would call it "clever", yes.

henryiii: For something that doesn't change often, a better solution would be to include the static information, then use the clever trick to verify the static information is not out of date. I'll make a PR with better mypy support.

eduardo-rodrigues: OK, sounds good.

BTW, we need to come back to a thread with Hans on how to populate the is_XXX and some other methods in `Particle`. It's very much related. I will ping you directly ... :-)

PR #275, opened 2020-11-09
feat: static typing

henryiii:
This adds full static typing, including for external users. Followup of sorts to #274.

- feat: better support for static types
- feat: type the rest of the code
- feat: support external type checking

### DIFF
--- a/src/particle/converters/bimap.py
+++ b/src/particle/converters/bimap.py
@@ -18,9 +18,29 @@ import csv
 from .. import data
 from ..exceptions import MatchingIDNotFound
 
+from typing import (
+    Union,
+    TextIO,
+    Callable,
+    TypeVar,
+    Generic,
+    overload,
+    Type,
+    Tuple,
+    Any,
+    Dict,
+    Iterator,
+)
 
-class BiMap(object):
+A = TypeVar("A")
+B = TypeVar("B")
+A_conv = Callable[[str], Union[A, int]]
+B_conv = Callable[[str], Union[B, int]]
+
+
+class BiMap(Generic[A, B]):
     def __init__(self, class_A, class_B, converters=(int, int), filename=None):
+        # type: (Type[A], Type[B], Tuple[A_conv, B_conv], Union[str, TextIO, None]) -> None
         """
         Bi-bidirectional map class.
 
@@ -61,21 +81,22 @@ class BiMap(object):
         >>> bimap = BiMap(PDGID, PythiaID, filename=filename)
         """
 
-        self.class_A = class_A
-        self.class_B = class_B
+        self.class_A = class_A  # type: Type[A]
+        self.class_B = class_B  # type: Type[B]
 
         name_A = self.class_A.__name__.upper()
         name_B = self.class_B.__name__.upper()
 
         if filename is None:
             filename = "{a}_to_{b}.csv".format(a=name_A.lower(), b=name_B.lower())
-            filename = data.open_text(data, filename)
-        elif not hasattr(filename, "read"):
+            file_object = data.open_text(data, filename)
+        elif hasattr(filename, "read"):
+            file_object = filename
+        else:
             # Conversion to handle pathlib on Python < 3.6:
-            filename = str(filename)
-            filename = open(filename)
+            file_object = open(str(filename))
 
-        with filename as _f:
+        with file_object as _f:
             self._to_map = {
                 converters[1](v[name_B]): converters[0](v[name_A])
                 for v in csv.DictReader(l for l in _f if not l.startswith("#"))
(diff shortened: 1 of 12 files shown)


src/particle/converters/bimap.py
```diff
@@ -61,21 +81,22 @@ def __init__(self, class_A, class_B, converters=(int, int), filename=None):
...
+        elif hasattr(filename, "read"):
+            file_object = filename
+        else:
             # Conversion to handle pathlib on Python < 3.6:
-            filename = str(filename)
-            filename = open(filename)
+            file_object = open(str(filename))
 
-        with filename as _f:
+        with file_object as _f:
```
henryiii, line 99: It's best to avoid changing types of a variable in the middle of a function. It also makes it easier for a reader to follow, generally.

src/particle/converters/bimap.py
```diff
@@ -212,16 +248,16 @@ def __getitem__(self, value):
...
 
     def __repr__(self):
+        # type: () -> str
         return "<{self.__class__.__name__}({a}->{b}): {n} matches>".format(
             self=self, a=self.name_A, b=self.name_B, n=self.__len__()
         )
 
-    def __str__(self):
-        return repr(self)
-
```
henryiii, line 224: This method is needed for int subclasses, because we have to replace the existing one. Otherwise, though, this is exactly the default definition! :)

src/particle/data/__init__.py
```diff
@@ -5,6 +5,6 @@
 # or https://github.com/scikit-hep/particle for details.
 
 try:
-    from importlib.resources import open_text
+    from importlib.resources import open_text  # type: ignore
 except ImportError:
-    from importlib_resources import open_text  # type: ignore
+    from importlib_resources import open_text
```
henryiii, line 10: This is messy. The standard library got `importlib.resources ` in 3.7 and `open_text` in 3.8 or something like that, so MyPy is not always happy. This appeases the pre-commit run of mypy, so that's enough for now.

src/particle/particle/convert.py
```diff
@@ -266,40 +279,40 @@ def get_from_pdg_mcd(filename):
...
-            "Mass",
-            "MassUpper",
-            "MassLower",
-            "Width",
-            "WidthUpper",
-            "WidthLower",
-            "NameCharge",
-        ),
-    )
+    with filter_file(filename) as file_object:
```
henryiii, line 282: We should always close files

src/particle/particle/convert.py
```diff
@@ -309,9 +322,9 @@ def get_from_pdg_mcd(filename):
         abcd = nc[1].str.split(",", 4, expand=True)
         d["charge"] = abcd[i]
         d.set_index("ID", inplace=True)
-        ds.append(d)
+        ds_list.append(d)
 
-    ds = pd.concat(ds)
+    ds = pd.concat(ds_list)
```
henryiii, line 327: Same comment about dynamically changing types part way through a function.

src/particle/particle/convert.py
```diff
@@ -386,10 +401,10 @@ def produce_files(particle2008, particle2020, version, year):
     with open(particle2020, "w", newline="\n", encoding="utf-8") as f:
         f.write(version_header(particle2020, version))
         new_table.to_csv(f, float_format="%.12g")
-    f.close()
```
henryiii, line 389: This is automatic with a context manager. Even better, the context manager will still close the file even if you error out! (which is why we never manually call `.close()`)

src/particle/particle/particle.py
```diff
@@ -585,7 +585,7 @@ def load_table(cls, filename=None, append=False, _name=None):
 
     # The following __le__ and __eq__ needed for total ordering (sort, etc)
 
-    def __le__(self, other):
+    def __lt__(self, other):
```
henryiii, line 588: Fixed a bug here! The definition was `<`, but the name as `<=`! MyPy expected `__lt__` to make it orderable, so it sort of caught the issue.

src/particle/pdgid/pdgid.py
```diff
@@ -63,8 +63,47 @@ def info(self):
...
+    three_charge = property(_functions.three_charge)
+
+
+# Verify the PDGID class has all relevant functions defined in the pdgid.functions module
 for _n in _fnames:
-    _decorator = property(getattr(_functions, _n), doc=getattr(_functions, _n).__doc__)
-    setattr(PDGID, _n, _decorator)
+    assert _n in dir(
+        PDGID
+    ), "{} missing from PDGID class! Update the list in pdgid.py".format(_n)
```
henryiii, line 109: This is much better statically, allowing "PDGID" to be checked and completed by static tools. We check these at run time and throw an error if it's not kept up (and this doesn't change that rapidly).

eduardo-rodrigues:
It's getting late on my side. I will look at this great PR tomorrow ...

src/particle/converters/bimap.py
```diff
@@ -18,9 +18,29 @@
...
+    Iterator,
+)
+
+A = TypeVar("A")
+B = TypeVar("B")
+A_conv = Callable[[str], Union[A, int]]
+B_conv = Callable[[str], Union[B, int]]
+
+
+class BiMap(Generic[A, B]):
```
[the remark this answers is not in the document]
eduardo-rodrigues: I believe you here!

src/particle/converters/bimap.py
```diff
@@ -18,9 +18,29 @@
...
 
-
-class BiMap(object):
+from typing import (
+    Union,
+    TextIO,
+    Callable,
+    TypeVar,
+    Generic,
+    overload,
```
eduardo-rodrigues, line 27: For looks, maybe better to put this decorator at the end of the import, since not quite the same as the other imports?

src/particle/converters/bimap.py
eduardo-rodrigues: Good point. Thanks for the improvement.

src/particle/converters/bimap.py
eduardo-rodrigues: Sure :-).

src/particle/particle/convert.py
eduardo-rodrigues: Yep. This was sloppy. Thanks.

src/particle/particle/convert.py
eduardo-rodrigues: Good to know - did not know in case of error.

src/particle/particle/particle.py
eduardo-rodrigues: Cool.

eduardo-rodrigues (approved):
This is great to have. Thank you!

src/particle/pdgid/pdgid.py
eduardo-rodrigues: Point taken.

src/particle/converters/bimap.py
henryiii: Sure, I can sort these, which will do it.

PR #278, opened 2020-11-12
Enhance Particle.to_list and to_dict

eduardo-rodrigues:
Add more filter functionality to `Particle.to_list` and `Particle.to_dict` since these 2 methods rely internally on `Particle.findall`, and the latter allows for more arguments than the former methods presently accepted. This PR makes the signatures of these methods match.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -65,6 +65,7 @@ test =
 dev =
     pytest
     pandas; python_version>"3.4"
+    tabulate
     check-manifest>=0.42
     black==20.8b1
     mypy==0.790
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,6 +61,7 @@ where=src
 test =
     pytest
     pandas; python_version>"3.4"
+    tabulate
 dev =
     pytest
     pandas; python_version>"3.4"
--- a/setup.cfg
+++ b/setup.cfg
@@ -72,6 +72,7 @@ dev =
 all =
     pytest
     pandas; python_version>"3.4"
+    tabulate
     check-manifest>=0.42
     black==20.8b1
     mypy==0.790
--- a/src/particle/particle/particle.py
+++ b/src/particle/particle/particle.py
@@ -285,6 +285,8 @@ class Particle(object):
         exclude_fields=(),  # type: Iterable[str]
         n_rows=-1,  # type: int
         filter_fn=None,  # type: Optional[Callable[[Particle], bool]]
+        particle=None,  # type: Optional[bool]
+        **search_terms  # type: Any
     ):
         # type: (...) -> List[List[Any]]
         """
@@ -328,6 +330,11 @@ class Particle(object):
         filter_fn: function, optional, default is None
             Apply a filter to each particle.
             See `findall(...)`` for typical use cases.
+        particle: bool, optional, default is None
+            Pass particle=True/False to only return particles or antiparticles.
+            Option passed on internally to `findall(...)``.
+        search_terms: keyword arguments, optional
+            See `findall(...)`` for typical use cases.
 
         Returns
         -------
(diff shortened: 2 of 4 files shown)


eduardo-rodrigues:
OK, the issue with pre-commit is that the most recent versions no longer support Python 3.5, see https://github.com/mgedmin/check-manifest/blob/master/CHANGES.rst. I will update the pre-commit checks so that it rather does 3.6 instead of 3.5 ...

eduardo-rodrigues:
OK, I quick on the pre-commit error. Maybe an issue from the fact that "pre-commit.ci is currently in alpha preview"? This is ready for review @henryiii.

henryiii:
Yes, I bet this is an issue with the pre-commit service. The pre-commit test we have in the CI works fine. I'll raise an issue with them, shouldn't block this. However, I have an idea that might help.

henryiii:
Actually, I have two ideas...

henryiii:
I don't know why the checks listed above still has a red x, the commit itself has a green checkmark and pre-commit is listed as passing...

setup.cfg
```diff
@@ -61,6 +61,7 @@ where=src
 test =
     pytest
     pandas; python_version>"3.4"
+    tabulate
```
henryiii, line 64: I'd probably list this in dev, too?

src/particle/particle/particle.py
```diff
@@ -285,6 +285,8 @@ def to_list(
         exclude_fields=(),  # type: Iterable[str]
         n_rows=-1,  # type: int
         filter_fn=None,  # type: Optional[Callable[[Particle], bool]]
+        particle=None,  # type: Optional[bool]
+        **search_terms  # type: Any
     ):
         # type: (...) -> List[List[Any]]
```
henryiii, line 291: It would be nice to have something more concrete here eventually. Like `List[Tuple[float, float, str, blah, blah]]`. Assuming the particle table has a strong structure to it?

setup.cfg
eduardo-rodrigues: Makes sense. Just added.

src/particle/particle/particle.py
eduardo-rodrigues: Yes. I would not know how mypy would say "potentially a list of arguments that can be bool, int, str, float" (probably that's it).  Can we leave this for now unless you know how to express the beast?

henryiii: Ahh, but it can change with the arguments. Okay, this is likely the best that we can do.

setup.cfg
henryiii: And all, just below, for sure!

henryiii (approved):
Good except for adding "tabulate" in a couple more extras.

src/particle/particle/particle.py
henryiii: Yes, a `ResType = Union[bool, int, str, float]` would be better than `Any`, but sadly you can't describe the ordering, since it depends on the arguments. But `List[List[ResType]` would be better (or just write it inline instead of making an alias).

setup.cfg
eduardo-rodrigues: Oops, damn .. for sure! (I am swinging between 2 totally different bits of work, and that's not good ;-)) Will commit in a sec.

src/particle/particle/particle.py
eduardo-rodrigues: You lost me here. Should I then replace `Any` with `List[List[Union[bool, int, str, float]]]`?

henryiii:
```suggestion
        # type: (...) -> List[List[Union[bool, int, str, float]]]
```
This is the simplest thing, though you could stick `Something = Union[bool, int, str, float]` somewhere in the file then use that instead of nesting it. But this is fine.

eduardo-rodrigues: OK, thanks.

eduardo-rodrigues: For reference, this caused trouble. I reverted back but will put it on my to-do list to look better at the issue, which comes from inside `to_dict` when calling `to_list` internally.

henryiii: I fixed it but you merged already. The problem was that you claimed this returned a dist with `List[str]` as a key, which is invalid, you can't have a mutable item as a key. Fixing the return type fixed the issue (plus a little cleanup).

eduardo-rodrigues: My apologies for that @henryiii.

PR #279, opened 2020-11-13
fix: type incorrect for return

henryiii:
Followup to #278

### DIFF
--- a/src/particle/particle/particle.py
+++ b/src/particle/particle/particle.py
@@ -288,7 +288,7 @@ class Particle(object):
         particle=None,  # type: Optional[bool]
         **search_terms  # type: Any
     ):
-        # type: (...) -> List[List[Any]]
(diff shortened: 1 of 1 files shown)


src/particle/particle/particle.py
```diff
@@ -534,9 +534,10 @@ def to_dict(cls, *args, **kwargs):
         """
         query_as_list = cls.to_list(*args, **kwargs)
 
-        return dict(
-            zip(query_as_list[0], zip(*query_as_list[1:]))
-        )  # dict(zip(keys, values))
```
henryiii, line 539: If you have to describe what you mean in a code-like comment - just write it that way in the code! :)

eduardo-rodrigues: Fair enough.

eduardo-rodrigues (approved):
Thank you for this fix!

PR #280, opened 2020-11-22
Add tests to check if every particle is parsed and added correctly

tamasgal:
I wanted to some cross-checks with the Julia package [Corpuscles.jl](https://github.com/JuliaPhysics/Corpuscles.jl) and noticed that some particles are not missing in all `particle*.csv` datasets after loading them. E.g. the currently default `particle2020.csv` contains a total of 610 particles, but only 603 are loaded:

```python
>>> from particle import Particle, data

>>> Particle.load_table(data.open_text(data, 'particle2020.csv'))

>>> len(Particle.findall())
603
```

I quickly created a diff to check which particles are missing compared to `Corpuscles.jl` and got:

```julia
julia> [Particle(p) for p in setdiff(Set(corpuscles_ids), Set(particle_ids))]
7-element Array{Particle,1}:
 Particle(-203338) Omega(2250)
 Particle(-103326) Xi(1950)
 Particle(-203316) Xi(2030)
 Particle(-103316) Xi(1950)
 Particle(-203326) Xi(2030)
 Particle(-203312) Xi(1690)
 Particle(-203322) Xi(1690)
```

A quick debug session in `particle` revealed that the problem is `Error: -5 is not a valid Parity` which is suppressed here: https://github.com/scikit-hep/particle/blob/4d58415cc3583192cf2e8df2a3881259fcbd806d/src/particle/particle/particle.py#L616 and originates from the enum `Parity`: https://github.com/scikit-hep/particle/blob/4d58415cc3583192cf2e8df2a3881259fcbd806d/src/particle/particle/enums.py#L34

There is already an enum entry for `Parity=5`, so this PR adds `-5`, named `x`.

### DIFF
--- a/src/particle/particle/enums.py
+++ b/src/particle/particle/enums.py
@@ -37,7 +37,6 @@ class Parity(IntEnum):
     p = 1
     m = -1
     u = 5
-    x = -5
 
 
 class Charge(IntEnum):
--- a/src/particle/particle/enums.py
+++ b/src/particle/particle/enums.py
@@ -37,6 +37,7 @@ class Parity(IntEnum):
     p = 1
     m = -1
     u = 5
+    x = -5
 
 
 class Charge(IntEnum):
--- a/tests/particle/test_particle.py
+++ b/tests/particle/test_particle.py
@@ -348,6 +348,21 @@ def test_explicit_table_loading():
     assert Particle.all() is not None
 
 
+def test_all_particles_are_loaded():
+    Particle.load_table(data.open_text(data, "particle2018.csv"))
+    assert len(Particle.all()) == 605
+    Particle.load_table(data.open_text(data, "particle2019.csv"))
+    assert len(Particle.all()) == 610
+    Particle.load_table(data.open_text(data, "particle2020.csv"))
+    assert len(Particle.all()) == 610
+
+    Particle.load_table(data.open_text(data, "nuclei2020.csv"))
+    assert len(Particle.all()) == 5880
+
+    # Load default table to restore global state
+    Particle.load_table()
+
+
 checklist_html_name = (
     (22, "&gamma;"),  # photon
     (1, "d"),  # d quark


eduardo-rodrigues:
Thanks as ever @tamasgal for being a great user, and for your careful checks.

I will look at all this asap. For now let me just say that Parity=-5 has no reason. It is kind of saying it's an undefined-bis. Indeed Parity can only be +/- 1 or undefined/irrelevant. Upfront I would not be surprised if the trouble is rather coming from a typo in the original 2008 PDG file with extended info, which never got updates after that one year of release. Let me get to this later today ...

eduardo-rodrigues:
Adding the check is of course excellent thing to have. I wonder how we managed to not added it till now :D.

tamasgal:
:see_no_evil: 

Yes, indeed, Parity other than +-1 is of course trouble but there are plenty of entries with `5`, so I thought it's a workaround and `-5` was forgotten and falls into the same category (undefined/irrelevant).

Anyways, let's see where it comes from... ;)

henryiii:
Let's fix a couple of things. We should a) throw an error instead of bypassing this (at the _very_ least in the tests, likely everywhere), abd b) compare lengths so we can tell if particles are missing in the tests. That way we keep the benefit of this excellent detective work in the future. :)

eduardo-rodrigues:
Got it! The issue was a bit more subtle than I thought at first, though I got hinted by the fact that the parity was reversed for some antibaryons only. So the issue is in the flip of parity for baryons -> antibaryons when producing the CSV data files! The flip of parity sign was done without an "if", and of course it should not be done for undefined parities :D. Outch. That's the definition of a(n introduced) bug.

(Note of course that you only found this issue for very badly experimentally known particles, hence this is not a problem
for the general user. Still.)

A proper fix is the following:
- Add the check for undefined parities when flipping values for antibaryons.
- Rebuild the CSV data files.
- Actually remove from the Parity enums the value 0, which is also not valid, since not a physical value.

I will put up a fix as soon as I have time - as you can see I'm multi-tasking today ;-). Then we should retrigger your PR. In any case, please remove that update to the enum, which isn't physical.

Thanks again.

henryiii:
I have a feeling that `value = int(v["ID"])` is actually the only thing that was supposed to be in that try/except. We don't want to hide errors after that. Basically, it should be, I think:

```
try:
    value = int(v["ID"])
except ValueError:
    continue
```

eduardo-rodrigues:
I will look into that as well, @henryiii. But in any case the issue I describe *is* a bug, and one that is physics-related, as it were.

henryiii:
Yes, I'm looking at something completely orthogonal; how do we catch and reduce these bugs in the future; I'm not looking into the actual bug at all since you have that. :)

eduardo-rodrigues:
Can you @tamasgal rebase on master to pick up my fixes?

tamasgal:
Yes, done!

eduardo-rodrigues:
Thanks.

PR #291, opened 2021-04-01
Add a .zenodo.json file

eduardo-rodrigues:
As per out chat, @matthewfeickert.

As you can see I'm not filling in all possible fields, e.g. the ones on versioning. I would say we go ahead assuming Zenodo will add the missing fields since anyway it *is* creating at present such a file, on the fly, from best-guesses, as it says on their site.

Let me know if you seen anything obvious missing. Otherwise approve :-). Advance thanks.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,0 +1,24 @@
+{
+    "description": "PDG particle data and identification codes",
+    "license": "other-open",
+    "upload_type": "software",
+    "creators": [
+        {
+            "affiliation": "University of Liverpool",
+            "name": "Eduardo Rodrigues",
+            "orcid": "0000-0003-2846-7625"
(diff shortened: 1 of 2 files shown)


matthewfeickert (approved):
LGTM. 

We (pyhf) originally had some difficulty in getting things to look right as the JSON schema seemed tempremental, but with help from the Zenodo team we got things settled out. From comparing this to the `pyhf` file and to the export of the (fully generated) JSON metadata for `particle` https://zenodo.org/record/4629613/export/json it seems that this should be good to go. You can also always go in and edit a release by hand.

eduardo-rodrigues:
Thanks a lot @matthewfeickert for having a look. Let's go with this, then.

PR #292, opened 2021-04-06
Programmatic name for diquarks and SUSY particles

eduardo-rodrigues:
The programmatic names of diquarks such as `(bs)(0)` where returned as `_bs_0`, where the prefix `_` is not so pretty. This PR introduces a trivial adaptation to rather return  `bs_0` and similarly for the other diquarks.

(OK, diquarks are not observed particles so this is largely academic/aesthetics.)

Some temporary variables were also exposed. Now that are removed, the right thing to do.

### DIFF
--- a/src/particle/particle/utilities.py
+++ b/src/particle/particle/utilities.py
@@ -33,7 +33,7 @@ def programmatic_name(name):
         .replace("+", "_plus")
     )
     # Strip off the ugly "_" at beginning of a name, such as when dealing with name="(bs)(0)""
-    return re.sub("^_", "", name)
+    return name.lstrip("_")
 
 
 def str_with_unc(value, upper, lower=None):
--- a/src/particle/particle/utilities.py
+++ b/src/particle/particle/utilities.py
@@ -19,7 +19,19 @@ def programmatic_name(name):
     name = re.sub("^~", "tilde_", name)
     # The remaining "~" now always means it's an antiparticle
     name = name if "~" not in name else "".join(name.split("~")) + "_bar"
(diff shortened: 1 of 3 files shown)


eduardo-rodrigues:
So now we have the following:
```
>>> programmatic_name("~chi(1)0")
'tilde_chi_1_0'  # was 'chi_1_0_bar', which was wrong
>>> programmatic_name("d(R)~")
'd_R_bar'
>>> programmatic_name("~d(R)~")
'tilde_d_R_bar'  # was also returning 'd_R_bar' as above, which was wrong
>>> programmatic_name("(bs)(0)")
'bs_0'  # was '_bs_0', which was a bit ugly
```
Much better :-).

eduardo-rodrigues:
I will merge this later today since no magic, just some reasonable improvements. I have another improvement in the pipeline, related.

src/particle/particle/utilities.py
```diff
@@ -29,6 +32,8 @@ def programmatic_name(name):
         .replace("-", "_minus")
         .replace("+", "_plus")
     )
+    # Strip off the ugly "_" at beginning of a name, such as when dealing with name="(bs)(0)""
+    return re.sub("^_", "", name)
```
henryiii, line 36:
```suggestion
    return name.lstrip("_")
```

henryiii (approved):

src/particle/particle/utilities.py
eduardo-rodrigues: Good idea. Indeed simpler.

PR #293, opened 2021-04-09
Particle and PDG ID literals for all particles in the latest data file

eduardo-rodrigues:
Seems the better thing to do :-).


eduardo-rodrigues:
@henryiii, drop a line if you have / do not have time to review this PR. I can proceed if you are overloaded.

PR #305, opened 2021-06-12
Bug fix for bug introduced in #279

eduardo-rodrigues:
Unfortunately the bug had been introduced trying to simplify the code, and no test was made for that :S.


eduardo-rodrigues:
Hi @henryiii, have you got any idea on how to solve this MyPy error? I'm lost in translation.

henryiii:
Probably, I can take a look in a hour or two.

src/particle/particle/particle.py
```diff
@@ -427,7 +429,7 @@ def to_list(
 
     @classmethod
     def to_dict(cls, *args, **kwargs):
-        # type: (Any, Any) -> Dict[str, List[Union[str, float, bool, int]]]
+        # type: (Any, Any) -> Dict[str, Tuple[Union[bool, int, str, float], ...]]
```
henryiii, line 431: This returns a tuple, not a list.

henryiii: And the key is a string, it's the value that's a Tuple.

eduardo-rodrigues: I had a Tuple before in my many trials but the hints were tricking me. For sure `str, Tuple` makes sense. I was definitely missing the `...`. Thanks a lot!

PR #315, opened 2021-08-12
PDG 2021 .mcd file - remove duplicate entry for the K*(892)+

eduardo-rodrigues:
The 2020 file does not have the issue.

The problem is that the PDG has 2 entries for the K*(892)+ mass, see https://pdglive.lbl.gov/Particle.action?init=0&node=M018&home=MXXX020.:
```
CHARGED ONLY, HADROPRODUCED  891.67±0.26891.67±0.26 MeV
CHARGED ONLY, PRODUCED IN τ LEPTON DECAYS  895.5±0.8895.5±0.8 MeV
```
and we obviously can't keep both in the dataframe.

I removed the line corresponding to "CHARGED ONLY, HADROPRODUCED" for consistency with the 2020 file.


eduardo-rodrigues:
Follow-up / update from PR https://github.com/scikit-hep/particle/pull/314.

PR #316, opened 2021-08-12
Make curated PDG particle data table 2021 the default

eduardo-rodrigues:
Following up from https://github.com/scikit-hep/particle/pull/314 and https://github.com/scikit-hep/particle/pull/315.

All data CSV files are updated to version 10 for consistency in the set.


eduardo-rodrigues:
FYI @henryiii and @HDembinski. I will merge this later today unless I hear anything against it.

For the next release we could still maybe address some of the issues you @HDembinski created yesterday since that breaks compatibility no matter which suggested option we take. I would make a version 0.16 (keeping 1.0 for bigger changes planned).

Issue #313, opened 2021-08-11
Suggestion to deprecate and later remove `Particle.find`


HDembinski:
Let me start again with my nitpick by saying that I like `particle` very much, I use all the time in my code.

I suggest to deprecate `Particle.find` and to remove it in a future version. It is fully superseded by `Particle.findall` which is the correct interface for a query (well, actually #312 is). `Particle.find` can only return one result, but that is a very special case when the query fundamentally can yield none or several entries. Having such a restricted return type is not a good design and should be a non-starter. The API works around this by raising exceptions whenever the API contract cannot be fulfilled, this works correctly, but it's duck-tape over the more fundamental issue.

Scott Meyers says: "Make interfaces easy to use correctly and hard (or impossible) to use incorrectly". `Particle.find` is very difficult to use correctly (in the sense that you do not get an exception), because a priori, the user cannot know whether the query will yield exactly one result. Frequently, the results will contain zero or more entries, but these states cannot be represented by the method.

If you are not convinced yet, there are three other closely-related design principles which `Particle.find` violates. 

1) Each function/method should do the absolute minimum amount of work that is necessary to respond to the user request.

`find` always first finds all the matches and stores them in a temporary list. But it could stop already when the query found two matches, because then the result is clear, an exception has to be thrown. Also storing everything in a list only to discard that list afterwards is wasting resources. The implemention only needs to store the first match and once a second mathc is found, it can abort by raising an exception. This could be fixed by implementing `finditer` as suggested in #312 and use that internally instead of `findall`, but it would be better to just remove this method.

2) Whenever feasible, fully return the result of some work. In other words, minimize the amount of temporary data that is generated and then thrown away internally in the implementation to respond to some user request.

Functions and methods should be so designed as to return the maximum amount of information for work that has been already done. The idea here is that the work has been performed already so it would be a waste to automatically discard some of that work. Let the user decide what to discard. Even if the user gets more than they intended to get, they may actually find that extra information useful, and if not, it is no problem for them to discard the extra information. If the disposal happens automatically, the extra amount of work is simply wasted.

According to this principle and since `find` first finds all the matches, it should also return all the matches. It already did the work to find all of them, the resources have been used. Designing the API in such a way that then only one result is returned means to automatically discard the rest of the work, without giving the user a choice to look at this extra information. Again, this could be fixed with #312, but `findall` is already the right kind of implementation which returns all results of the query.

3) Only raise exceptions in exceptional situations.

[...]

Issue #312, opened 2021-08-11
Add version of Particle.findall which returns generator?


HDembinski:
I love the API of `Particle.findall`. A minor nitpick for me is that it returns a list, while the modern way would be to return a generator.

I think it would be easy to change the implementation to return a generator. However, simply replacing the return type in this method with a generator would be a breaking change. So perhaps another method could be added that returns the generator, and `Particle.findall` internally just calls that. I am not sure how this should be called, but perhaps `Particle.finditer`?

henryiii:
It is extremely important that the list retuned by `findall` is sorted. The "first" results are always the simplest; DecayLangauge uses this. However, we could make sure the original list of particles is sorted; then returning an iterator from `finditer` would be possible; as a bonus, this might be a bit faster too after the initial setup, since it doesn't have to sort after finding particles.

henryiii:
(The other option would be for `findall` to return `sorted(finditer)`, but it would be much nicer to not have the ordering different between the two)

eduardo-rodrigues:
> It is extremely important that the list retuned by `findall` is sorted. The "first" results are always the simplest; DecayLangauge uses this. However, we could make sure the original list of particles is sorted; then returning an iterator from `finditer` would be possible; as a bonus, this might be a bit faster too after the initial setup, since it doesn't have to sort after finding particles.

I like it too that the first results are the most relevant. As far as DecayLanguage is concerned the package no longer relies on ordering or findall, in fact. This is because DecayLanguage needs by construction to deal transparently with particles in the PDG *and many* other particles that are only known to EvtGen, newly defined particles, exotic beasts, etc. A while back I made the searches and "translations needed robust by using directly the bimaps I defined in Particle.

In short, we have no constraints from that side.

eduardo-rodrigues:
Now, since we are discussing the API in non-backward compatible ways together with the suggestion to make `find` obsoleted, see https://github.com/scikit-hep/particle/issues/313 - if we go the way of providing only a method, eventually, then one could actually go the following way:
- Keep `findall` as-is for now.
- Break the API and have `find` return a generator for all search results.
- then remove findall since `find` would do as `findall` and then the distinction in name isn't needed anymore.

The one downside of this is that `find` would be totally different.

Note that the one different as of now between `findall` and `find` is that the latter throws an exception if more than one particle is found in the search. This we can get rid of if people are happy to check for the number of returned particles.

Let me know what you think. I dumped these thoughts fast so may have missed a point.

[...]

PR #318, opened 2021-08-14
Deprecate Particle.find

eduardo-rodrigues:
Closes https://github.com/scikit-hep/particle/issues/313.


eduardo-rodrigues:
Following our discussions at https://github.com/scikit-hep/particle/issues/313 and https://github.com/scikit-hep/particle/issues/312, @HDembinski, @henryiii and @mayou36. Let me know if you would see anything missing for this first step discussed. I would separate things to would address https://github.com/scikit-hep/particle/issues/312 separately in other PRs; seems best.

eduardo-rodrigues:
FYI @HDembinski, @henryiii and @mayou36 I will merge this later today since it seems just want we converge.

.pre-commit-config.yaml
```diff
@@ -29,4 +29,4 @@ repos:
   hooks:
   - id: mypy
     files: src
-    additional_dependencies: [attrs==19.3.0]
+    additional_dependencies: [attrs==19.3.0, types-Deprecated]
```
henryiii, line 32:
```suggestion
    additional_dependencies: [attrs==19.3.0, types-deprecated]
```

Ultra minor, but packages are usually lower case, even though mypy suggests the cased version. I sometimes see "Cython" instead of "cython", but usually it's lower case everywhere.

henryiii (approved):

.pre-commit-config.yaml
eduardo-rodrigues: Fair enough. I got this example online, actually.

henryiii:
PS: I can turn off Sourcery on Particle/DecayLanguage - it's nice to enable, run, then disable once in a while; that's what I'm doing on my repos. The comment-on-very-push is too irritating for me. Fine if you like it, but if you don't, I can turn it off.

henryiii:
(Codecov is also doing that, so you might be fine with it. :) )

eduardo-rodrigues:
> PS: I can turn off Sourcery on Particle/DecayLanguage - it's nice to enable, run, then disable once in a while; that's what I'm doing on my repos. The comment-on-very-push is too irritating for me. Fine if you like it, but if you don't, I can turn it off.

I find it irritating as well but have been OK since the packages are not under heavy development.

Issue #299, opened 2021-05-19
Put coverage tests back online


eduardo-rodrigues:
Our good old coverage tests aren't active anymore, see the link https://dev.azure.com/scikit-hep/particle/_build/latest?definitionId=1?branchName=master from the README. We should have *some coverage testing* back up and running.

eduardo-rodrigues:
Hi @matthewfeickert, @henryiii, @mayou36, are you performing coverage tests in your packages with GHA? I used to have that with Azure but now went on with GHA totally; as a result coverage tests are temporarily gone and I would appreciate suggestions on how to get them back in GHA the best way. Advance thanks for any pointers.

mayou36:
We're using the coverage package (and still the old way with report). There is a GHA available (https://github.com/marketplace/actions/coveralls-github-action) that I would suggest to use.

eduardo-rodrigues:
> We're using the coverage package (and still the old way with report). There is a GHA available (https://github.com/marketplace/actions/coveralls-github-action) that I would suggest to use (but actually not tried yet myself).

Hey, so I gave this a try with PR https://github.com/scikit-hep/particle/pull/319 but I don't get it to work! Looking around at various issues and discussions online it seems to me that the app does not work for us because it releases on lcov.info. As others on the web I wasted quite some time and finally gave it a try with Codecov (https://github.com/marketplace/actions/codecov), and that seems to work, see again https://github.com/scikit-hep/particle/pull/319. I think I'm going with that instead!

eduardo-rodrigues:
Closed via https://github.com/scikit-hep/particle/pull/319.

henryiii:
I use coveralls in at least one package, but I've never had much success with the action - I've skipped most of the issues by just using the uploader directly. You just have had to (or still have to) specify "github" as the service to get it to work properly with GITHUB_TOKEN.  See https://github.com/tomerfiliba/plumbum/blob/3876425a04f934667e1f91ba88f526ca6976dbd8/.github/workflows/ci.yml#L81

Both are quite fine, though; happy to see either service being used. If we can get a nice comparison at some point, we could add one of them to the developer guidelines. Or maybe add both.

eduardo-rodrigues:
Thanks for the pointer. Yes, test coverage is something not yet ... covered in our guidelines. Let us get a bit more experience with the tools and then proceed with a paragraph of info ... The summaries with Covedoc look good to me.

PR #319, opened 2021-08-17
Get coverage back and switch to Codecov on GHAs

eduardo-rodrigues:
After lots of trials and time wasted, see below, it seems that https://github.com/marketplace/actions/coveralls-github-action does not work. Rather switching to Codecov, and it works just fine!


eduardo-rodrigues:
Closes https://github.com/scikit-hep/particle/issues/299.

henryiii:
This looks nice and clean. 👍

PR #321, opened 2021-08-17
chore: use newer files interface

henryiii:
Addressing https://github.com/scikit-hep/decaylanguage/issues/138 (but for particle).

A few other bits of cleanup. Adding a noxfile too to make it easier to run.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -249,8 +249,8 @@ If you want a non-default data file distributed with the package just proceed as
 .. code-block:: python
 
     >>> from particle import data
-    >>> Particle.load_table(data.open_text(data, 'particle2018.csv'))
-    >>> Particle.load_table(data.open_text(data, "nuclei2020.csv"), append=True)  # I still want nuclei info
+    >>> Particle.load_table(data.files / "particle2018.csv"))
+    >>> Particle.load_table(data.files / "nuclei2020.csv"), append=True)  # I still want nuclei info
     >>> Particle.table_names()  # list the loaded tables
 
 
--- a/README.rst
+++ b/README.rst
@@ -249,8 +249,8 @@ If you want a non-default data file distributed with the package just proceed as
 .. code-block:: python
 
     >>> from particle import data
-    >>> Particle.load_table(data.files / "particle2018.csv"))
-    >>> Particle.load_table(data.files / "nuclei2020.csv"), append=True)  # I still want nuclei info
+    >>> Particle.load_table(data.basepath / "particle2018.csv"))
+    >>> Particle.load_table(data.basepath / "nuclei2020.csv"), append=True)  # I still want nuclei info
     >>> Particle.table_names()  # list the loaded tables
 
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,8 @@ python_requires = >=2.7, !=3.0, !=3.1, !=3.2, !=3.3, !=3.4
 install_requires =
     enum34>=1.1; python_version<"3.4"
     typing>=3.7; python_version<"3.5"
-    importlib_resources>=1.0; python_version<"3.7"
+    typing_extensions; python_version<"3.8"
+    importlib_resources>=2.0; python_version<"3.9"
     attrs>=19.2
     hepunits>=2.0.0
     deprecated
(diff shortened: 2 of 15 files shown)


setup.cfg
```diff
@@ -60,24 +60,20 @@ where=src
...
     tabulate
 dev =
-    pytest
+    check-manifest>=0.42
     pandas; python_version>"3.4"
+    pytest
     tabulate
-    check-manifest>=0.42
-    black==20.8b1
-    mypy==0.790
```
henryiii, line 73: It's invalid to run this by hand, we should only support running from pre-commit; otherwise we'd have to maintain these with correct versions, etc.

src/particle/converters/bimap.py
```diff
@@ -89,12 +90,13 @@ def __init__(self, class_A, class_B, converters=(int, int), filename=None):
...
             filename = "{a}_to_{b}.csv".format(a=name_A.lower(), b=name_B.lower())
-            file_object = data.open_text(data, filename)
-        elif hasattr(filename, "read"):
+            file_object = data.files.joinpath(filename).open()
+        elif isinstance(filename, HasRead):
             file_object = filename
+        elif isinstance(filename, HasOpen):
+            file_object = filename.open()
         else:
-            # Conversion to handle pathlib on Python < 3.6:
```
henryiii, line 96: These workarounds were fixed by using a Protocol.

setup.cfg
eduardo-rodrigues: I didn't write this ;-).

henryiii: Didn't imply you did. :) I was originally trying to support both, but obviously pre-commit was (automatically) maintained and this one was forgotten.

tests/particle/test_particle.py
```diff
@@ -23,15 +23,16 @@
 
 def test_find():
     # 1 match found
-    prepr = repr(Particle.find(name="gamma"))
+    with pytest.deprecated_call():
+        prepr = repr(Particle.find(name="gamma"))
```
henryiii, line 27: You should always check for warnings (with `pytest.warns`) and deprecation calls. PyTest logs should be clean, and you do want to ensure these are producing the expected warnings. Technically, this is probably producing the wrong type of deprecation call - a user facing deprecation call should use `category=FutureWarning`, but that can be fixed in a followup PR.

eduardo-rodrigues: Thanks. Makes sense. Indeed the tests were producing warnings. I actually fixed a couple of genuine `FutureWarning` recently.

README.rst
```diff
@@ -249,8 +249,8 @@ If you want a non-default data file distributed with the package just proceed as
 .. code-block:: python
 
     >>> from particle import data
-    >>> Particle.load_table(data.open_text(data, 'particle2018.csv'))
-    >>> Particle.load_table(data.open_text(data, "nuclei2020.csv"), append=True)  # I still want nuclei info
+    >>> Particle.load_table(data.files / "particle2018.csv"))
```
eduardo-rodrigues, line 252: Below I see some other example usage like `filename = data.files("pdgid_to_pythiaid.csv")` so wonder if both this one and the above one with a slash are viable? In any case `Particle.load_table(data.files / "particle2018.csv"))` seems to have a closing parenthesis too many.

src/particle/converters/bimap.py
```diff
@@ -175,7 +177,7 @@ def DirectionalMaps(name_A, name_B, converters=(str, str), filename=None):
     --------
 
     >>> from particle import data  # doctest: +SKIP
-    >>> filename = data.open_text(data, 'a_to_b.csv')  # doctest: +SKIP
+    >>> filename = data.files / "a_to_b.csv"  # doctest: +SKIP
```
eduardo-rodrigues, line 180: Not `filename = data.files("a_to_b.csv")`?

src/particle/data/__init__.py
```diff
@@ -4,7 +4,19 @@
...
+if sys.version_info < (3, 9):
+    import importlib_resources as resources
+else:
+    import importlib.resources as resources
+
+
+files = resources.files(__name__)
+
+
+open_text = deprecated(version="0.13.0", reason="Use particle.data.files instead.")(
```
eduardo-rodrigues, line 20:
```suggestion
open_text = deprecated(version="0.16.0", reason="Use particle.data.files instead.")(
```

eduardo-rodrigues (changes requested):
Nice 👍 .

README.rst
henryiii: `filename = data.files("pdgid_to_pythiaid.csv")` Oops, that would be a typo. `data.files` is basically a Path, so it's not callable - it should be either `data.files / name` or the identical, more verbose, but more chain able `data.files.joinpath(name)`.

BTW, is "files" a good name for this? We could call it `data.basepath` or something else like that.

src/particle/data/__init__.py
henryiii: Thanks! I thought I fixed this, noticed my mistake later.

README.rst
eduardo-rodrigues: Good that this was caught. Yeah, I like `data.basepath` more. Seems clearer 👍.

eduardo-rodrigues (approved):

PR #322, opened 2021-08-18
Add check for duplicate entries in .mcd PDG data files

eduardo-rodrigues:
Indeed it has happened already in 2019 and 2021 that the file contained duplicate entries, see https://github.com/scikit-hep/particle/pull/315 for the 2021 case. Better have a the `convert.py` script augmented to check for such issues.


tests/particle/test_generation.py
```diff
@@ -43,7 +41,7 @@ def test_generate(tmp_path):
 
 
 @pytest.mark.parametrize("filename", FILES)
-def test_file_dup(filename):
+def test__csv_file_duplicates(filename):
```
eduardo-rodrigues, line 44:
```suggestion
def test_csv_file_duplicates(filename):
```

tests/particle/test_convert.py
```diff
@@ -0,0 +1,24 @@
...
+# or https://github.com/scikit-hep/particle for details.
+
+try:
+    from pathlib2 import Path
+except ImportError:
+    from pathlib import Path
+
+import pytest
+
+from particle import data
```
eduardo-rodrigues, line 14:
```suggestion
```

eduardo-rodrigues:
OK, ready for review @henryiii.

tests/data/README.rst
```diff
@@ -0,0 +1,5 @@
+Particle test data folder contents
+==================================
+
+This folder contains files used for testing purposes.
+By construction files can have features, hence not be fully correct.
```
henryiii, line 5: What is this line trying to say?

henryiii (approved):

tests/data/README.rst
eduardo-rodrigues: Really only that in case somebody looks at the files for examples. Better make it clear that they are not examples to follow. The file is not super important but I thought it does not harm.

PR #324, opened 2021-08-20
New PDGID related functions

eduardo-rodrigues:
Quite a few things here:

- New functions `is_sm_lepton`, `is_sm_quark` and `is_excited_quark_or_lepton`. These provide the necessary "coverage".
- `is_composite_quark_or_lepton` deprecated in favour of better named `is_excited_quark_or_lepton`.
- Improved docstrings for `is_quark` and `is_lepton` and these related new functions.
- All PDGID related functions and the `PDGID` class actually accessible with a `from particle.pdgid import *`.
- `is_lepton` fixed to match the behaviour of the related `is_quark`, i.e. now excited leptons are not considered leptons (only SM and 4th generation leptons).
- Test suite enhanced accordingly.


src/particle/pdgid/functions.py
```diff
@@ -99,12 +101,37 @@ def is_quark(pdgid):
...
 
 
+def is_sm_quark(pdgid):
+    # type: (PDGID_TYPE) -> bool
+    """
+    Does this PDG ID correspond to a Standard Model quark?
+
+    Codes 1-8 are reserved for Standard Model or fourth-generation quarks,
+    (excluding excited, i.e. composite, quarks)
+    but only the codes 1-6 actually correspond to SM quarks.
```
jonas-eschle, line 111:
```suggestion
	Fourth-generation leptons are not included.
    Codes 1-8 are reserved for Standard Model or fourth-generation quarks,
    (excluding excited, i.e. composite, quarks)
    but only the codes 1-6 actually correspond to SM quarks.
```

Maybe make it clearer that they are not "sm"?

src/particle/pdgid/functions.py
```diff
@@ -99,12 +101,37 @@ def is_quark(pdgid):
...
+    """
+    return 11 <= abspid(pdgid) <= 18
+
+
+def is_sm_lepton(pdgid):
+    # type: (PDGID_TYPE) -> bool
+    """
+    Does this PDG ID correspond to a Standard Model lepton?
+
+    Codes 11-18 are reserved for Standard Model or fourth-generation leptons,
```
jonas-eschle, line 131:
```suggestion
	Fourth-generation leptons are not included.
    Codes 11-18 are reserved for Standard Model or fourth-generation leptons,
```
Same as above

jonas-eschle (approved):
Looks good to me, maybe clarify the "SM" formulation?

src/particle/pdgid/functions.py
```diff
@@ -444,6 +471,20 @@ def is_composite_quark_or_lepton(pdgid):
     return _digit(pdgid, Location.N) == 4 and _digit(pdgid, Location.Nr) == 0
 
 
+@_deprecated(
+    version="0.16.0",
+    reason="This method is deprecated and will be removed from version 0.17.0. Use is_excited_quark_or_lepton instead.",
```
henryiii, line 476:
```suggestion
    reason="This method will be removed from version 0.17.0. Use is_excited_quark_or_lepton instead.",
```

This already states "deprecated" in the warning, no need to duplicate.

tests/pdgid/test_functions.py
```diff
@@ -130,6 +130,22 @@ def test_is_quark(PDGIDs):
...
+        PDGIDs.DQuark,
+        PDGIDs.UQuark,
+        PDGIDs.SQuark,
+        PDGIDs.CQuark,
+        PDGIDs.BQuark,
+        PDGIDs.TQuark,
+    )
+    _non_sm_quarks = [id for id in PDGIDs if id not in _sm_quarks]
+    for id in _sm_quarks:
+        assert is_sm_quark(id) == True
```
henryiii, line 144:
```suggestion
        assert is_sm_quark(id)
```

Never use `==` with the literals `True` and `False`. Either use `is` if you really want to make sure that it returns a bool, or just check directly.

tests/pdgid/test_functions.py
```diff
@@ -130,6 +130,22 @@ def test_is_quark(PDGIDs):
...
+    for id in _non_sm_quarks:
+        assert is_sm_quark(id) == False
```
henryiii, line 146:
```suggestion
        assert not is_sm_quark(id)
```

tests/pdgid/test_functions.py
henryiii: (same below in other additions - I'm also fixing existing ones based on Flake8 reports in a branch)

henryiii (commented):
Fine with me other than the existing comments, once addressed consider my review to be "Approve". :)

tests/pdgid/test_functions.py
eduardo-rodrigues: Good catch - this is a copy and paste from other similar code but it did not trigger my eye.

src/particle/pdgid/functions.py
eduardo-rodrigues: @jonas-eschle, the suggestion indicates the very same text or it is my eyes :S?

jonas-eschle: There is a "Fourth-generation leptons are not included." in the beginning. I thought it may could be a bit clearer that this function does not return the fourth generation. The docs says about which codes are for what, but not which codes the function actually returns (kinda, implicitly, by explicitly saying "SM or fourth-generation".

jonas-eschle: When reading it, I just didn't directly realise what the function now actually returns, although technically it is written. Just a bit more explicitly could be good, but maybe it's just me.

So also like this it's fine. My suggestion would be to just make it a bit clearer what the "sm" means here

eduardo-rodrigues: OK, will reread and commit a little update. Thanks for the comments - docs are really important :-).

henryiii (approved):

src/particle/pdgid/functions.py
eduardo-rodrigues: Done with my docstring update. I will now proceed and merge since Henry has a clean-up MR in the pipeline. Feel 100% free to comment further or suggestion other updates :-). Thanks again.

PR #326, opened 2021-08-20
style: make changes suggested by flake8

henryiii:
Will put a few comments inline.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -81,3 +81,8 @@ testpaths =
     tests
 junit_family = xunit2
 addopts = '-Wd'
+
+[flake8]
+max-complexity = 24
+ignore = E203, E231, E501, E722, W503, B950, E402
+select = C,E,F,W,B,B9
--- a/setup.cfg
+++ b/setup.cfg
@@ -68,7 +68,7 @@ test =
     pytest
     pytest-cov
     tabulate
-    pandas;python_version>"3.4"
+    pandas;python_version>"3.4" and python_version<"3.10"
 
 [options.package_data]
 * = *.csv, *.fwf, *.mcd, *.py.typed
--- a/src/particle/__init__.py
+++ b/src/particle/__init__.py
@@ -5,23 +5,15 @@
 # or https://github.com/scikit-hep/particle for details.
 
 import sys
+from typing import Tuple
 
 # Direct access to other ID classes
 from .geant import Geant3ID
 
 # Direct access to Particle literals
 # Direct access to Particle (the CSV file is not read until a particle is accessed)
-from .particle import (
-    Charge,
-    Inv,
-    InvalidParticle,
-    Parity,
-    Particle,
-    ParticleNotFound,
-    SpinType,
-    Status,
-    literals,
-)
+from .particle import InvalidParticle, Particle, ParticleNotFound, literals
+from .particle.enums import Inv, Parity, SpinType, Status
 
 # Direct access to PDGID
 from .pdgid import PDGID
@@ -38,3 +30,27 @@ from .converters import Pythia2PDGIDBiMap
 # Direct access to handy LaTeX to HTML particle name conversions
 # Direct access to kinematics functions
 from .particle import latex_to_html_name, lifetime_to_width, width_to_lifetime
+
+__all__ = (
+    "Geant3ID",
+    "Charge",
+    "Inv",
+    "InvalidParticle",
+    "Parity",
+    "Particle",
+    "ParticleNotFound",
+    "PythiaID",
+    "PDGID",
+    "Pythia2PDGIDBiMap",
+    "latex_to_html_name",
+    "lifetime_to_width",
+    "width_to_lifetime",
+    "Status",
+    "SpinType",
+    "__version__",
+)
+
+
+def __dir__():
+    # type: () -> Tuple[str, ...]
+    return __all__
--- a/src/particle/__init__.py
+++ b/src/particle/__init__.py
@@ -34,21 +34,21 @@ from .converters import Pythia2PDGIDBiMap
 from .particle import latex_to_html_name, lifetime_to_width, width_to_lifetime
 
 __all__ = (
-    "Geant3ID",
     "Charge",
+    "Geant3ID",
     "Inv",
     "InvalidParticle",
+    "PDGID",
     "Parity",
     "Particle",
     "ParticleNotFound",
-    "PythiaID",
-    "PDGID",
     "Pythia2PDGIDBiMap",
+    "PythiaID",
+    "SpinType",
+    "Status",
     "latex_to_html_name",
     "lifetime_to_width",
     "width_to_lifetime",
-    "Status",
-    "SpinType",
     "__version__",
 )
 
--- a/src/particle/__init__.py
+++ b/src/particle/__init__.py
@@ -4,6 +4,8 @@
 # Distributed under the 3-clause BSD license, see accompanying file LICENSE
 # or https://github.com/scikit-hep/particle for details.
 
+from __future__ import absolute_import
+
 import sys
 from typing import Tuple
 
@@ -13,7 +15,7 @@ from .geant import Geant3ID
 # Direct access to Particle literals
 # Direct access to Particle (the CSV file is not read until a particle is accessed)
 from .particle import InvalidParticle, Particle, ParticleNotFound, literals
-from .particle.enums import Inv, Parity, SpinType, Status
+from .particle.enums import Charge, Inv, Parity, SpinType, Status
 
 # Direct access to PDGID
 from .pdgid import PDGID
--- a/src/particle/particle/convert.py
+++ b/src/particle/particle/convert.py
@@ -53,7 +53,17 @@ When you are done, you can save one or more of the tables:
 
 import os
 from datetime import date
-from typing import Any, Callable, Dict, Iterable, List, Optional, TextIO, TypeVar, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    Optional,
+    TextIO,
+    Tuple,
+    TypeVar,
+    Union,
+)
 
 import numpy as np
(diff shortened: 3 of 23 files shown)


src/particle/__init__.py
```diff
@@ -38,3 +30,27 @@
...
+    "width_to_lifetime",
+    "Status",
+    "SpinType",
+    "__version__",
+)
+
+
+def __dir__():
+    # type: () -> Tuple[str, ...]
+    return __all__
```
henryiii, line 56: These should be in most "public" files, and should help tab completion be more precise. Anything not in `__all__` will not show up in tab completion (Python 3.7+ only). @eduardo-rodrigues please check `__all__` and make sure I didn't miss anything, or add something that should not be public. It should be pretty easy, since most things in `__init__.py` files are either in `__all__` or have a `noqa` since they are unused (in which case they are basically hidden).

src/particle/particle/convert.py
```diff
@@ -75,13 +85,24 @@
     Inv_mapping,
     Parity,
     Parity_mapping,
-    SpinType,
-    Status,
```
henryiii, line 79: These were unused. Assuming users should not do `particle.particle.SpinType`, but either `particle.SpinType` or `particle.enums.SpinType`.

src/particle/particle/particle.py
```diff
@@ -402,7 +408,7 @@ def to_list(
             for fld in exclude_fields:
                 try:
                     tbl_names.remove(fld)
-                except:
+                except ValueError:
```
henryiii, line 411: No blank excepts! Catches memory errors, Ctrl-C, etc...

src/particle/particle/particle.py
```diff
@@ -564,7 +570,7 @@ def load_table(cls, filename=None, append=False, _name=None):
                 cls.load_table(f, append=True, _name="nuclei2020.csv")
             return
         elif isinstance(filename, HasRead):
-            tmp_name = _name or getattr(filename, "name")
+            tmp_name = _name or filename.name
```
henryiii, line 573: Not any safer than the simple expression. Unless we intended to have it "fall-back" on an empty string or something (in which chase it needs the three argument form).

src/particle/pdgid/__init__.py
```diff
@@ -61,6 +61,15 @@
...
 
+from typing import Tuple
+
+from .functions import is_composite_quark_or_lepton  # noqa: F401
+from .functions import is_gauge_boson_or_higgs  # noqa: F401
+from .functions import is_generator_specific  # noqa: F401
+from .functions import is_quark  # noqa: F401
+from .functions import is_sm_gauge_boson_or_higgs  # noqa: F401
+from .functions import is_special_particle  # noqa: F401
+from .functions import is_technicolor  # noqa: F401
```
henryiii, line 72: Are all these really supposed to be missing from `__all__`? `is_composite_quark_or_lepton` I assume yes, but `is_quark`, for example?

tests/converters/test_maps.py
```diff
@@ -20,7 +20,7 @@ def test_BiMap():
     assert "BiMap(PDGID-PythiaID)" in str(bimap)
 
     with pytest.raises(MatchingIDNotFound):
-        pyid = bimap[PDGID(9000221)]
+        bimap[PDGID(9000221)]
```
henryiii, line 23: Assignment does nothing.

tests/particle/test_generation.py
```diff
@@ -36,8 +36,8 @@ def test_generate(tmp_path):
 
     particle2021_data = data.basepath / "particle2021.csv"
     with particle2021.open() as src, particle2021_data.open() as res:
-        src = [l for l in src.readlines() if not l.startswith("#")]
-        res = [l for l in res.readlines() if not l.startswith("#")]
+        src = [x for x in src.readlines() if not x.startswith("#")]
+        res = [x for x in res.readlines() if not x.startswith("#")]
```
henryiii, line 40:
```suggestion
        src = [line for line in src.readlines() if not line.startswith("#")]
        res = [line for line in res.readlines() if not line.startswith("#")]
```

This was the first one I changed, I think line is better than x. `l` is ambiguous with `1` sometimes (and `I`, but we need that one).

tests/particle/test_particle.py
```diff
@@ -43,11 +43,15 @@ def test_lambda_style_search():
     assert 1000010010 in particles
     assert -1000010010 in particles
 
-    [p.pdgid for p in Particle.findall(lambda p: p.pdg_name == "p" and p > 0)] == [
+    assert [
+        p.pdgid for p in Particle.findall(lambda p: p.pdg_name == "p" and p > 0)
+    ] == [
```
henryiii, line 48: This originally did nothing, was a pointless comparison (and several following it).

henryiii:
Tests are currently taking ~17 minutes on Python 3.10 because we are building wheels for NumPy because we have to build the Pandas wheel for 3.10, and it pins an old NumPy during the build. I'm looking at turning it off for 3.10 tests at the moment.

eduardo-rodrigues:
> Tests are currently taking ~17 minutes on Python 3.10 because we are building wheels for NumPy because we have to build the Pandas wheel for 3.10, and it pins an old NumPy during the build. I'm looking at turning it off for 3.10 tests at the moment.

Yeah, sounds good. I don't think there is much gain for now in doing 3.10

src/particle/particle/particle.py
eduardo-rodrigues: Hmm, we had been through these before. Damn, it creeped in again!

src/particle/particle/convert.py
eduardo-rodrigues: Safe to remove for this script. In fact I removed a few such unused imports also in PR https://github.com/scikit-hep/particle/pull/324 but did not run a comprehensive check. Good that you did it 👍.

setup.cfg
```diff
@@ -68,7 +68,7 @@ test =
     pytest
     pytest-cov
     tabulate
-    pandas;python_version>"3.4"
+    pandas;python_version>"3.4" and python_version<"3.10"
```
henryiii, line 71: Check now takes 54 seconds instead of 18 minutes. :)

eduardo-rodrigues: Cool 👍! (But of course that time gets a lot down after getting up a lot with the on-the-spot builds for 3.10 ;-).)

src/particle/__init__.py
eduardo-rodrigues: This I did not know. Thank you for the info. OK, will look at that for DecayLanguage - extra on my to-do list.

src/particle/pdgid/__init__.py
eduardo-rodrigues: I fixed a bunch of these when adding several new PDGID related functions in https://github.com/scikit-hep/particle/pull/324. If you could rebase on master as soon as that is merged some will be gone for sure. I will then check carefully what exactly we want public ... Advance thanks.

tests/converters/test_maps.py
eduardo-rodrigues: That's right. I thought not assigning to a tmp variable would rather be bad practice (or else assign to `_`). But fine as you changed since the instantiation is what really matters.

src/particle/pdgid/__init__.py
henryiii: Sure, will do!

src/particle/particle/particle.py
henryiii: Thats why they need to be checked in CI (as there are now!) :)

src/particle/particle/particle.py
eduardo-rodrigues: No. Your update seems fine.

tests/particle/test_particle.py
eduardo-rodrigues: Ah, the `assert` was missing :S!

src/particle/pdgid/__init__.py
eduardo-rodrigues: You can now rebase and then I will have a review for what should be public or private ...

henryiii: Done.

src/particle/__init__.py
```diff
@@ -38,3 +32,27 @@
 # Direct access to handy LaTeX to HTML particle name conversions
 # Direct access to kinematics functions
 from .particle import latex_to_html_name, lifetime_to_width, width_to_lifetime
+
+__all__ = (
+    "Geant3ID",
```
eduardo-rodrigues, line 37: Funny that these do not all get sorted. Almost, but not quite. No big deal.

src/particle/pdgid/__init__.py
```diff
@@ -61,6 +61,15 @@
 
 from __future__ import absolute_import
 
+from typing import Tuple
+
+from .functions import is_composite_quark_or_lepton  # noqa: F401
```
eduardo-rodrigues, line 66: All these should be imported normally as all others, no noqa.

henryiii: Then they need to be added to `__all__`, they were missing before.

src/particle/pdgid/__init__.py
```diff
@@ -106,6 +108,8 @@
 from .pdgid import PDGID
 
 __all__ = (
+    "PDGID",
```
eduardo-rodrigues, line 111: I had added PDGID to the list below (you got it when rebasing). This is now a duplicate.

henryiii: Okay!

eduardo-rodrigues (approved):
Just a couple of tiny comments otherwise LGTM 👍.

src/particle/__init__.py
henryiii: Only the actual imports are sorted, the `__all__` remains as-is (though I can manually sort, looks nicer that way).

PR #329, opened 2021-08-26
feat: add finditer

### LINKED ISSUE #312: Add version of Particle.findall which returns generator?

henryiii:
Closes #312. Containership testing will be slower, but findall will likely be faster (not that anything is very slow).

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -5,7 +5,7 @@ Version 0.16.0 (in preparation)
 --------------
 
 - `Particle` class:
-  - Added `finditer`, which return an iterator instead of a complete list like `findall`.
+  - Added `finditer`, which returns an iterator instead of a complete list like `findall`.
   - Method `find` deprecated. It will be removed from version 0.17.0 onwards.
     Please use the more general `findall` method, or the new `finditer` method.
 - Data files:
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -5,8 +5,9 @@ Version 0.16.0 (in preparation)
 --------------
 
 - `Particle` class:
-  - Method `find`deprecated. It will be removed from version 0.17.0 onwards.
-    Please use the more general `findall` method.
+  - Added `finditer`, which return an iterator instead of a complete list like `findall`.
+  - Method `find` deprecated. It will be removed from version 0.17.0 onwards.
+    Please use the more general `findall` method, or the new `finditer` method.
 - Data files:
   - Added the 2021 PDG data table .mcd file, now default.
   - Version 10 of package CSV data files:
--- a/src/particle/particle/particle.py
+++ b/src/particle/particle/particle.py
@@ -1180,7 +1180,6 @@ C (charge parity) = {C:<6}  I (isospin)       = {self.I!s:<7}  G (G-parity)     
         >>> # Trivially find all pseudoscalar charm mesons
         >>> Particle.findall(lambda p: p.pdgid.is_meson and p.pdgid.has_charm and p.spin_type==SpinType.PseudoScalar)  # doctest: +SKIP
 
-        See also ``find``, which throws an exception if the particle is not found or too many are found.
         See also ``finditer``, which provides an iterator instead of a list.
         """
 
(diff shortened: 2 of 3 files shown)


docs/CHANGELOG.md
```diff
@@ -5,8 +5,9 @@ Version 0.16.0 (in preparation)
 --------------
 
 - `Particle` class:
-  - Method `find`deprecated. It will be removed from version 0.17.0 onwards.
-    Please use the more general `findall` method.
+  - Added `finditer`, which return an iterator instead of a complete list like `findall`.
```
eduardo-rodrigues, line 8:
```suggestion
  - Added `finditer`, which returns an iterator instead of a complete list like `findall`.
```

eduardo-rodrigues (approved):
👍

eduardo-rodrigues:
Morning @HDembinski and @jonas-eschle. This MR from @henryiii introduces `Particle.finditer` and is about to get merged. It deals with the suggestion and discussions at https://github.com/scikit-hep/particle/issues/312. Any comment or this is totally ready to go?

jonas-eschle:
Seems reasonable to me, it's explicit.

*(Another note, just slightly related: since the `find` is deprecated, it's maybe worth to point also to alternatives, such as the `from_...` with names methods, as they are (AFAIK) not mentioned anywhere (except of the `from_pdgid`). But that's for another PR (just to avoid that people use the `find('name')[0]`)*

eduardo-rodrigues:
> Seems reasonable to me, it's explicit.
> 
> _(Another note, I forgot it in the deprecation, just slightly related: since the `find` is deprecated, it's maybe worth to point also to alternatives, such as the `from_...` with names methods, as they are (AFAIK) not mentioned anywhere (except of the `from_pdgid`). But that's for another PR (just to avoid that people use the `find('name')[0]`)_

We can address further docs and hints in a separate PR, sure. If you can give that a thought as an "outsider" that would be great :-). Thanks.

HDembinski:
finditer looks fine, great.
Is findall returning a sorted list or not? The implementation looks like it is not sorted, or maybe I am missing something. I think that the raison d'etre for findall is that it sorts the results, if that is not happening, then it could also be removed if you want to be minimalist, but I think it is still reasonable to keep both for convenience.

henryiii:
@HDembinski the internal particle table is now sorted, so the return from both finditer and findall is sorted by construction, since it finds the first match, then the next, etc. `list(Particle.finditer())` is identical, but I think it's nice to keep both, as particle was designed to be comfortable and easy from the command line, and wrapping stuff in list is mildly irritating. And probably more importantly, `re.finditer` and `re.findall` is a common example of something with both, and has influenced our design. Finally, removing every previous way to find particles with new functions that only exist in the current version is unnecessarily forcing everyone to adapt and not support old versions. Currently, we can just ask users to use `.findall` if they need to support all Particle versions.

@jonas-eschle 
```
(particle,) = Particle.finditer(...) # or findall
```

Is identical to the old behavior, that will throw an error if less than one or more than one particle was found.

jonas-eschle:
> ```python
> (particle,) = Particle.finditer(...) # or findall
> ```
> 
> Is identical to the old behavior, that will throw an error if less than one or more than one particle was found.

Yes, in a way: the errormessage is not very verbose, but better than just taking `[0]` of the findall, indeed.

It bears another disadvantage though (the same as `find` had): it's not future proof. If particles are added that may fulfill the criterion (be it, for the sake of argument, as we discover that a particle has different properties than currently assumed). Therefore (as mentioned in #331), it can be useful to have (and point to) a method that gives you an exact match (and not all as in `findall`)

> Something worth mentioning in the README and docstring?

I would, as I agree with @henryiii that it is technically truly "the old behavior"

HDembinski:
> ... and wrapping stuff in list is mildly irritating

True.

PR #334, opened 2021-09-10
Fix unreasonably slow import time

chrisburr:
Importing particle 0.16.0 takes an unreasonably long time:

```python
In [1]: %time import particle
CPU times: user 12.2 s, sys: 85.2 ms, total: 12.3 s
Wall time: 12.5 s
```

This is due to the use of `if value in cls._table:` inside the `Particle.load_table` which causes it be O(n²) with `Particle.__eq__` currently being called 21,095,760 million times.

This PR does the most backwards compatible fix possible by using a `set` to reduce this down to O(n):

```python
In [1]: %time import particle
CPU times: user 189 ms, sys: 13.7 ms, total: 202 ms
Wall time: 212 ms
```

IMHO a better solution might be to change `Particle._table` to be a `set` but this would change the public API and result in the ordering being lost.

### DIFF
--- a/src/particle/particle/particle.py
+++ b/src/particle/particle/particle.py
@@ -585,37 +585,40 @@ class Particle(object):
         with open_file as f:
             r = csv.DictReader(line for line in f if not line.startswith("#"))
 
+            # Use a set to avoid making this O(n^2)
+            known_particles = set(cls._table)
             for v in r:
                 try:
                     value = int(v["ID"])
 
                     # Replace the previous value if it exists
                     # We can remove an int; ignore typing thinking we need a particle
-                    if value in cls._table:
+                    if value in known_particles:
                         cls._table.remove(value)  # type: ignore
+                        known_particles.remove(value)  # type: ignore
 
-                    cls._table.append(
-                        cls(
-                            pdgid=value,
-                            mass=float(v["Mass"]),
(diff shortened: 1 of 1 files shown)


eduardo-rodrigues:
Thanks for reporting. This went unnoticed indeed. Hmm, it is the second time that this issue gets in, hence we should be more careful (maybe add a check on timing to load the default tables!). In the past the fix was trivial but this MR adds a few changes. I'm unlikely to be able to review before Monday ...

henryiii:
This was changed from a set to a list to allow findall to be faster in 0.16, since it no longer needs to sort the results of findall and therefore allows finditer to work without pre-computing and sorting. But, yes, this needs to be fixed.

henryiii:
We could simplify this solution slightly I think by just filling the set, then saving it as a list from sorted (which is done already).

henryiii:
Okay, this solution should be exactly as efficient as 0.16 everywhere except using a set instead of a list, removing that import issue. Thanks!

henryiii:
PS: you can use `-X importtime` or [PYTHONPROFILEIMPORTTIME](https://docs.python.org/3.10/using/cmdline.html#envvar-PYTHONPROFILEIMPORTTIME) to track module imports (Python 3.7+).

PR #336, opened 2021-09-10
chore: renable missing Python 3.5 support

henryiii:
This was only missing due to setup-cfg-fmt's default 3.6+, I believe.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["2.7", "3.6", "3.8", "3.9", "3.10-dev"]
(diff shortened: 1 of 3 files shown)


eduardo-rodrigues:
If you take care of this do not forget to follow-up with DecayLanguage, cf. our chat from 10-ish days ago. Then they are "in sync". Thank you.

eduardo-rodrigues (approved):

henryiii:
DecayLanguage can be more strict than particle - it's not add support for 3.5 just because Particle has it. Though I do intend to restore it. :) DecayLanguage could and possibly should drop 2.7 & 3.5 before Particle. I just don't see any reason to drop 3.5 without dropping 2.7. Dropping 3.5 does not give us anything at all, because we are compatible with 2.7, and most third party libraries have dropped these versions at the same time, too.

PR #337, opened 2021-09-12
style: enforce black on notebooks

henryiii:
1-2 lines might be a tiny bit worse, but having Black on the notebooks is really nice too.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
 - repo: https://github.com/psf/black
   rev: 21.8b0
   hooks:
-  - id: black
+  - id: black-jupyter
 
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -25,7 +25,7 @@ repos:
 - repo: https://github.com/psf/black
   rev: 21.8b0
   hooks:
-  - id: black
+  - id: black-jupyter
```
eduardo-rodrigues, line 28: So this does Black for everything as before *and* notebooks? Then cool 👍.

eduardo-rodrigues (approved):

.pre-commit-config.yaml
henryiii: Yep.

PR #347, opened 2021-10-28
add LHCb names

admorris:
Following a discussion I had with @eduardo-rodrigues on Monday, here is my attempt at adding LHCb particle names (as defined in [`ParticleTable.txt`](https://gitlab.cern.ch/lhcb-conddb/DDDB/-/blob/master/param/ParticleTable.txt)). For the most part they are either PDG or EvtGen names, but there are some which are neither. Several examples of these are included in the test.

The change involves quite a bit of code-duplication, so it might be worth a refactor to avoid this, particularly if we want to extend to other (experiment-specific?) naming schemes in the future.

Here is how I updated `conversions.csv`: https://gist.github.com/admorris/e1bd683e9ab5cd2a26c781f375a1dac2

The documentation hints at the existence of scripts for generating the `x_to_y.csv` files from `conversions.csv`, but I don't know where they are, so I generated `pdgid_to_lhcbname.csv ` myself and updated the comment in the other files.

PR #351, opened 2021-11-03
Do not expose converters in top-level imports

eduardo-rodrigues:
Following from the discussions at https://github.com/scikit-hep/particle/pull/347. Indeed it does not quite make sense to expose at top-level a single converter map from the set of a few defined by default.

Documentation will follow up in the README to expand the usage of converters.


eduardo-rodrigues:
In fact the README is ready as far as this change is concerned because it provides as import example `from particle.converters import Pythia2PDGIDBiMap`.

eduardo-rodrigues:
Merging since trivial :-).

Issue #331, opened 2021-08-30
Docs for methods `from_[string]`


jonas-eschle:
As far as I can see in the docs, there is no real mention of the methods that create a particle from an exact string match. This can be quite useful if the name of the particle is exactly known and is a better way to create it instead of calling `findall(...)[0]` (as the latter would not give an error if misspelled but could return (a list of) wrong particles).

Also, since the find is deprecated, it's maybe worth to point in the deprecation to these alternatives. Depending on the usecase (if you have the properties of a particle you go for `find{all,iter}`, if you have the exact name, you use `from_evtgen_name` or similar, as the `from_string` is deprecated I think?).

eduardo-rodrigues:
For reference we decided to deprecate `Particle.from_string()` and `Particle.from_string_list()`, see discussion under https://github.com/scikit-hep/particle/pull/352. A separate MR will be created ...

PR #352, opened 2021-11-04
Better coverage of documentation for Particle methods

eduardo-rodrigues:
Closes https://github.com/scikit-hep/particle/issues/331

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -136,28 +136,21 @@ or, say, the name used in EvtGen, you can get a particle directly.
     >>>
     >>> Particle.from_evtgen_name("J/psi")
     <Particle: name="J/psi(1S)", pdgid=443, mass=3096.900 ± 0.006 MeV>
-    >>>
-    >>> next(Particle.finditer('pi'))  # iterator of particles
-    <Particle: name="pi0", pdgid=111, mass=134.9768 ± 0.0005 MeV>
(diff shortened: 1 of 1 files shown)


eduardo-rodrigues:
Hey @jonas-eschle, I am here addressing https://github.com/scikit-hep/particle/issues/331 and a bit more.

I'm on the fence for `from_string()` and `from_string_list()`, which `from_string()` calls internally. In fact I would suggest that we deprecated them right now. For a start, `Particle.findall(pdg_name="pi")` is equivalent to `Particle.from_string_list("pi")` and admittedly a nicer and clearer call. Then, `from_string()` itself is special because it returns the "best match" from the list found by  `from_string_list()`, but "best match" is not necessarily a clearly defined definition for the user.

Happy for your thoughts but I reckon I will deprecate these, in which case the documentation will be adapted accordingly.

I need to add more on `finditer()` and that's the only reason why is this marked as Draft.

jonas-eschle:
> but "best match" is not necessarily a clearly defined definition for the user.

I agree, I think this is pretty much the conclusion that we also had about the "find best match" of the method `find`.

Deprecation seems reasonable to me

eduardo-rodrigues:
Ready for review. Feel free to merge if/when happy.

PR #354, opened 2021-11-09
Deprecate Particle methods from_string and from_string_list

eduardo-rodrigues:
Follow-up from discussion at https://github.com/scikit-hep/particle/pull/352#issuecomment-961138684.

### DIFF
--- a/src/particle/particle/particle.py
+++ b/src/particle/particle/particle.py
@@ -1238,7 +1238,11 @@ C (charge parity) = {C:<6}  I (isospin)       = {self.I!s:<7}  G (G-parity)     
         )
 
     @classmethod
-    def from_string(cls: type[Self], name: str) -> Self:
+    @deprecated(
+        version="0.21",
+        reason="This method is deprecated and will be removed from version 0.22.0. Use finditer or findall instead.",
+    )
+    def from_string(cls: type[Self], name: str) -> list[Self]:
         "Get a particle from a PDG style name - returns the best match."
         matches = cls.from_string_list(name)
         if matches:
@@ -1246,6 +1250,10 @@ C (charge parity) = {C:<6}  I (isospin)       = {self.I!s:<7}  G (G-parity)     
         raise ParticleNotFound(f"{name} not found in particle table")
 
     @classmethod
+    @deprecated(
+        version="0.21",
+        reason="This method is deprecated and will be removed from version 0.22.0. Use finditer or findall instead.",
+    )
     def from_string_list(cls: type[Self], name: str) -> list[Self]:
         "Get a list of particles from a PDG style name."
 
--- a/src/particle/particle/particle.py
+++ b/src/particle/particle/particle.py
@@ -12,6 +12,7 @@ import csv
 from copy import copy
 from functools import total_ordering
 from typing import Any, Callable, Iterable, Iterator, Sequence, SupportsInt, TypeVar
+from deprecated import deprecated
 
 # External dependencies
 import attr
@@ -1242,7 +1243,7 @@ C (charge parity) = {C:<6}  I (isospin)       = {self.I!s:<7}  G (G-parity)     
         version="0.21",
         reason="This method is deprecated and will be removed from version 0.22.0. Use finditer or findall instead.",
     )
-    def from_string(cls: type[Self], name: str) -> list[Self]:
+    def from_string(cls: type[Self], name: str) -> Self:
         "Get a particle from a PDG style name - returns the best match."
         matches = cls.from_string_list(name)
         if matches:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,7 @@ repos:
   hooks:
   - id: mypy
     files: src
-    additional_dependencies: [attrs==21.4.0, hepunits>=2.2.0, importlib_resources]
+    additional_dependencies: [attrs==21.4.0, hepunits>=2.2.0, importlib_resources, types-deprecated]
     args: [--show-error-codes]
 
 - repo: https://github.com/codespell-project/codespell
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ dependencies = [
     "hepunits>=2.0.0",
     "importlib-resources>=2.0;python_version<\"3.9\"",
     "typing-extensions;python_version<\"3.8\"",
+    "deprecated"
 ]
 dynamic = ["version"]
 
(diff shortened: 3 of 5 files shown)


eduardo-rodrigues:
OK @henryiii, @jonas-eschle, this is almost all that I was suggesting to deprecate. The remaining failures have got to do with AmpGen style of names that unfortunately do not follow any reasonable convention and seem largely ad-hoc for many particles. Terrible. Given that AmpGen (https://github.com/GooFit/AmpGen) is used by hardly nobody I woud suggest that we continue with this MR and remove the test related to AmpGen, rather making an issue for the repository to follow a reasonable convention for particle names; it could be the PDG, or EvtGen or simply a programmatic name as provided by `Particle`. This is my take and it makes sense to me, pragmatically speaking. Else we can't deprecate this and/or need to have some workarounds for no big reason (AmpGen).

eduardo-rodrigues:
Note that what we *could add* is a method `Particle.from_name`, which would have a similar handy role as  `Particle.from_pdgid`. Of course both searches are a simple single-particle match that can be done via the more general search methods `Particle.findall` and `Particle.finditer`; still, they are super handy and most often used.

eduardo-rodrigues:
Hi both, will you have time to look into this MR this week? If it helps I can make a separate MR for the suggested method `Particle.from_name` as that "non-disruptive" unlike a little thing in this MR.

src/particle/particle/particle.py
```diff
@@ -1289,6 +1297,10 @@ def from_string_list(cls, name):
             return []
 
     @classmethod
+    @deprecated(
+        version="0.16.0",
+        reason="This method is deprecated and will be removed from version 0.17.0. Use finditer or findall instead.",
+    )
     def _from_group_dict_list(cls, mat):
```
jonas-eschle, line 1304: No one should use this anyway, right? It's private, so not sure if a deprecation is needed? Maybe it's confusing if you see it as a user

henryiii:
I'll try to look into this soon, but what is the path forward for decaylanguage if this is removed? Half of decaylanuage (and literally the thing that gives it it's name) is parsing the AmpGen language, and that's why this was added in the first place.

src/particle/particle/particle.py
eduardo-rodrigues: You're totally write. I was "trigger happy" it seems. Rolling this back.

eduardo-rodrigues:
> I'll try to look into this soon, but what is the path forward for decaylanguage if this is removed? Half of decaylanuage (and literally the thing that gives it it's name) is parsing the AmpGen language, and that's why this was added in the first place.

Hi @henryiii, I would like to finalise a few long-standing matters in the package. This is the only outstanding MR.

I stand by my position that these unsafe and "fuzzy" methods should really be removed, see my comments at the top. This argument of yours is rather weak at this point, for several reasons:

- These days most of DecayLanguage is *not* related anymore with AmpGen. The package is massively about dec files and the classes I added, plus visualisation. The origins got obsoleted, as it were.
- I know you gave DecayLanguage its name but it does suit perfectly what the packages does overall and what just about anyone uses it for, including Belle (II). Can you point me to who is using this part I see as obsoleted here?
- In any case, can you point me to the places that actually use `Particle.from_string` or `Particle.from_string_list`, which is to say what code out there is using DecayLanguage's `modeling/amplitudechain.py`? I looked at GooFit and AmpGen but did not find any usage.

In short, I think we should move on and not stay anchored on some bit that is not being used anymore and is not ideal anyway, see my comment on the really bad choice of convention of particle names in AmpGen above.

Looking forward to your replies.

jonas-eschle:
> I'll try to look into this soon, but what is the path forward for decaylanguage if this is removed? Half of decaylanuage (and literally the thing that gives it it's name) is parsing the AmpGen language, and that's why this was added in the first place.

What about the other way around?  Changing things in AmpGen moving forward (Disclaimer, I am not very familiar with the AmpGen semantics)?

eduardo-rodrigues:
Thanks @jonas-eschle. Indeed I suggest that too above in the thread. In short, improve where it makes sense and avoid sub-optimal solutions or workarounds.

henryiii:
Quick update: Mike just submitted an NSF proposal that would work on fitting packages, and DecayLanguage (as a AmpGen parser) is included in that proposed work. If funded, it would supply the needed support to restore the AmpGen -> GooFit parser tests, and add other backends. I think should wait until DecayLanguage adapts/changes before stripping something from Particle that is needed for it to work. Once DecayLanguage can be shown to work without these methods, then they can be deprecated.

eduardo-rodrigues:
> Quick update: Mike just submitted an NSF proposal that would work on fitting packages, and DecayLanguage (as a AmpGen parser) is included in that proposed work. If funded, it would supply the needed support to restore the AmpGen -> GooFit parser tests, and add other backends.

That is good news for you folks. But do keep in mind one of my original comments - also from Jonas - that AmpGen should really use standard names and not cooked-up ones. Keeping fuzzy name matching for something that is no good and also hardly used by anybody does not make sense IMO. You still did not reply to me on who is using that stuff in DecayLanguage ...

> I think should wait until DecayLanguage adapts/changes before stripping something from Particle that is needed for it to work. Once DecayLanguage can be shown to work without these methods, then they can be deprecated.

Sure, we can adapt that part rather easily using standard particles names ;-).

Let's "reconvene" in the new year. Thanks.

eduardo-rodrigues:
Hello @henryiii, @jonas-eschle, as you may have seen, discussions are ongoing with AmpGen authors ...

In the meantime I rebased this rather old MR. One pragmatic solution at this point would be to:
- Get this merged  to flag deprecation no matter what.
- Move the 2 methods to the only place where a "translation" to home-made-ish particle names are needed, meaning DecayLanguage.
- Continue trying to convince AmpGen to move to stardar PDG names.

This way Particle becomes cleaner and no user of the functionality looses ability to continue to work. I really want to avoid fuzzy searches, as mentioned above, as they are not helping anyone in the end -  a user wants to search and find a particle given a name, not a likely match, since for the latter the powerful searches by properties are the way to go.

jonas-eschle:
I do agree, it signals the way forward for everyone,  IMHO, there is no need to keep up legacy methods without any warning to at least make it to the average user clear that this will change.

eduardo-rodrigues:
OK, finally created the complementary MR in `DecayLanguage`, see https://github.com/scikit-hep/decaylanguage/pull/312. Will move forward, also with discussions with AmpGen maintainers to converge on namings.

PR #356, opened 2021-11-11
chore: prepare for 0.16.2

eduardo-rodrigues:


eduardo-rodrigues:
FYI @henryiii, I am making this minor release because we may have to make a  release 0.17 if we do what I am suggesting in MR https://github.com/scikit-hep/particle/pull/354. Of course it is then best to have also the latest and greatest version just before the MR goes in.

PR #359, opened 2021-11-16
Add new class method Particle.from_name

eduardo-rodrigues:
Following the discussion at https://github.com/scikit-hep/particle/pull/354 and my suggestion, @henryiii and @jonas-eschle. Main reason is that we have `from_pdgid` and `from_evtgen_name` and it seemed awkward not to have `from_name`, the particle name itself.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Version 0.16.3 (WIP)
+--------------
+
+- `Particle` class:
+  - New class method `Particle.from_name`.
+
 
 Version 0.16.2
 --------------
--- a/src/particle/particle/particle.py
+++ b/src/particle/particle/particle.py
@@ -1054,11 +1054,13 @@ C (charge parity) = {C:<6}  I (isospin)       = {self.I!s:<7}  G (G-parity)     
             If no particle matches the input name.
         """
         try:
-            (particle,) = Particle.finditer(name=name)  # throws an error if < 1 or > 1 particle is found
+            (particle,) = Particle.finditer(
+                name=name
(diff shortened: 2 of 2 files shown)


eduardo-rodrigues:
Let me know @henryiii and @jonas-eschle if you have any thoughts against this addition to the API, which to me seems justified.

jonas-eschle (approved):
Sorry, I thought I had commented, but it got lost in a browser tab :)

Looks good to me, it's basically what we discussed can be done now using finditer to obtain exactly one matching particle.

There is one implementation comment, this try and catch: I am aware that it is more used in Python and I think there are great applications for this. But there is one large caveat: it's not really explicit and may catches more than expected if `finditer` raises an *actual ValueError*, say because of a bug inside it. This is then hidden. I think I would personally use a length check, but it's a minor detail/comment, it's also fine like this for me!

eduardo-rodrigues:
> Sorry, I thought I had commented, but it got lost in a browser tab :)

:D

> Looks good to me, it's basically what we discussed can be done now using finditer to obtain exactly one matching particle, just wrapped in a user-friendly function, nice!

Exactly. So now we have a better coverage to instantiate a single matched particle.

> There is one implementation comment, this try and catch: I am aware that it is more used in Python and I think there are great applications for this. But there is one large caveat: it's not really explicit and may catches more than expected if `finditer` raises an _actual ValueError_, say because of a bug inside it. This is then hidden. I think I would personally use a length check, but it's a minor detail/comment, it's also fine like this for me!

I'm not sure what would be best to have. For sure I have it on my to-dos to relook at test coverage at some point and then there should be a (admittedly )trivial test added for `from_name`. If `finditer` was to become buggy the test would fail.

PR #363, opened 2021-12-10
feat: drop Python 2

henryiii:
With Uproot and Awkward dropping Python 2, I think it's time to do it here too!

### DIFF
--- a/src/particle/particle/particle.py
+++ b/src/particle/particle/particle.py
@@ -1078,6 +1078,7 @@ C (charge parity) = {C:<6}  I (isospin)       = {self.I!s:<7}  G (G-parity)     
     def finditer(
         cls,
         filter_fn=None,  # type: Optional[Callable[[Particle], bool]]
+        *,
         particle=None,  # type: Optional[bool]
         **search_terms,  # type: Any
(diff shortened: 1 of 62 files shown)


src/particle/particle/particle.py
```diff
@@ -1124,9 +1121,6 @@ def finditer(
         See also ``findall``, which returns the same thing, but as a list.
         """
 
-        # Note that particle can be called by position to keep compatibility with Python 2, but that behavior should
-        # not be used and will be removed when support for Python 2.7 is dropped.
-
```
henryiii, line 1129: I've implemented this now, as promised.

eduardo-rodrigues (approved):
This all looks good to me. Thank you 👍!

src/particle/particle/particle.py
eduardo-rodrigues: To what are you refering to?

henryiii: `finditer(function, True)` is no longer allowed, now it must be `finditer(function, particle=True)`. Though I missed `findall` just below. Will fix.

Issue #361, opened 2021-12-02
HTML names: improve compatibility with other XML dialects (e.g. SVG) by only using unicode code-points


admorris:
In [the ntuple wizard](https://gitlab.cern.ch/lhcb-dpa/wp6-analysis-preservation-and-open-data/lhcb-ntuple-wizard-frontend/-/merge_requests/11) I found myself rendering particle HTML names in `viz.js` using some inline SVG. However, the named HTML entities (e.g. `&omega;`) had to be replaced with code-points (e.g. `&#x3c9;`) for SVG to understand them. According to [some guy on stackoverflow](https://stackoverflow.com/questions/18473312/how-do-i-display-html-encoded-values-in-svg/18473406#18473406) the latter are globally compatible with XML derivatives.

eduardo-rodrigues:
Hi @admorris, I missed somehow this issue and only saw it yesterday by change. Sorry for that.

The issue you are seeing is just with Greek letters or more generally with all special characters as for example in a complex particle name such as
```
>>> Particle.from_pdgid(-425).html_name
'D&#773;<SUB>2</SUB><SUP>*</SUP>(2460)<SUP>0</SUP>'
```

Sorting out the problem for all special characters is far more than just making the replacements for Greek letters.

admorris:
The broader issue is that named (rather than numbered) entities are specific to HTML, e.g. `&nbsp;` should become `&#160;` or `&#xa0;` to be supported by other XML dialects. In your example `&#773;` is fine. As far as I can see in the function `latex_to_html_name`, the only named entities happen to be Greek letters. If there is a function to convert a unicode character into its code-point in hexadecimal or decimal, then the diff should be minimal: we can do `greek_letter_name_to_unicode` then convert to the code-point.

```diff
--- a/src/particle/particle/utilities.py
+++ b/src/particle/particle/utilities.py
@@ -178,7 +178,9 @@ def latex_to_html_name(name):
     name = re.sub(r"\\mathrm\{(.*?)\}", r"\1", name)
     name = re.sub(r"\\left\[(.*?)\\right\]", r"[\1] ", name)
     for gl in _list_name_greek_letters:
-        name = name.replace(r"\%s" % gl, "&%s;" % gl)
+        # Get the hexadecimal representation of the unicode code-point
+        ucp = greek_letter_name_to_unicode(gl).encode("unicode_escape")[2:].decode("utf-8")
[...]

PR #366, opened 2021-12-17
HTML Greek letters to use hex code rather than entity name

eduardo-rodrigues:
Closes https://github.com/scikit-hep/particle/issues/361.

### DIFF
--- a/src/particle/particle/utilities.py
+++ b/src/particle/particle/utilities.py
@@ -6,6 +6,7 @@
 import math
 import re
 import unicodedata
+from html.entities import name2codepoint
 from typing import Optional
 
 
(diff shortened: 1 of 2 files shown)


eduardo-rodrigues:
See here @admorris for the improvement to better display HTML particle names with Greek letters in SVG.

eduardo-rodrigues:
Of course many tests fail because I pushed prior to getting the tests updated. I will update those asap ...

eduardo-rodrigues:
FYI @henryiii I will merge this in before the hols. Maybe you're still around and can have a look at this short update?

henryiii:
Seems good to me!

PR #374, opened 2022-01-10
Removal of deprecated methods and functions

eduardo-rodrigues:
As advertised with the 0.16 series.

Note that I'm not yet removing the `deprecated`module from the dependencies given the outstanding MR https://github.com/scikit-hep/particle/pull/354. Will get back to it as soon as the first Python 3 only release 0.20.0 is out.


eduardo-rodrigues:
(Doing this in parallel with other work made me commit several iterations. But now all ready for review.)

eduardo-rodrigues:
Can one of you look into this today, @henryiii and/or @jonas-eschle? Just in case you spot anything worth doing here. I would like to release 0.20 asap (step jump from 0.17 to mark the difference) to then move with other developments.

src/particle/particle/particle.py
```diff
@@ -1226,38 +1225,6 @@ def findall(
             cls.finditer(filter_fn=filter_fn, particle=particle, **search_terms)
         )
 
-    @classmethod
```
jonas-eschle, line 1229: While it is good to remove the legacy method, I also like to make it an explicit error in a first round (just a RuntimeError which says it was removed).

The motivation for this is that users tend to ignore deprecation warnings in my experience (developers don't). So the transition doesn't happen when it's deprecated, but when it fails. And an error saying it was removed is more user-friendly than a "method not found"

(I see a function raising a RuntimeError as "removed", physically removing it can still happen any time, it's already not "there" anymore)

but It's a bit of a style, however you prefer. I can imagine a lot of folks using this function somewhere (with non-updated code) and being surprised when it fails without a clear instruction on how to update.

jonas-eschle (commented):
Good! One comment, but other than that looks good to me

henryiii:
I'll check it in about an hour!

src/particle/particle/particle.py
eduardo-rodrigues: Oh, I see. I did not think about it. Have no strong opinion. What you suggest certainly makes very much sense for a package with a large user base. Here I dunno.

You mean something like
```
@classmethod
    def find(cls, *args, **search_terms):
        # type: (Any, Any) -> Particle
        raise RuntimeError("Method has been removed post versions 0.16. Please use finditer or findall instead.")
```
?

henryiii: You are reminding me that we need to move type comments into annotations. :)

That would work, no strong opinion, but I do see where it might be slightly nicer, prompting users in the right direction. I'd use `AttributeError`, though, so it doesn't modify behavior.

henryiii (approved):
Feel free to return a placeholder method as described, otherwise this looks good. I'll probably do a little bit of stylistic cleanup (type comments to type annotations) before 0.20 after this is merged.

src/particle/particle/particle.py
eduardo-rodrigues: OK, will make this kind of update tomorrow ...

eduardo-rodrigues:
> Feel free to return a placeholder method as described, otherwise this looks good. I'll probably do a little bit of stylistic cleanup (type comments to type annotations) before 0.20 after this is merged.

I would prefer that we first make the "major" release and your stylistic improvements go in a patch version as that does not affect functionality. I think this makes more sense.

src/particle/particle/particle.py
jonas-eschle: Yep exactly, and indeed, the AttributeError could make sense. As said, not cruciol, however you feel like

jonas-eschle (approved):

henryiii:
> I would prefer that we first make the "major" release and your stylistic improvements go in a patch version as that does not affect functionality. I think this makes more sense.

This is just a further cleanup that is part of the removal of Python 2 support - a followup for #363 - which seems reasonable to include in the release that drops Python 2. It also improves support for downstream type checking - mypy has started making type comments optional and not enabled by default (I think). It's also fully automated, so I can make it as soon as this PR goes in.

eduardo-rodrigues:
> This is just a further cleanup that is part of the removal of Python 2 support - a followup for #363 - which seems reasonable to include in the release that drops Python 2. It also improves support for downstream type checking - mypy has started making type comments optional and not enabled by default (I think). It's also fully automated, so I can make it as soon as this PR goes in.

OK, fair enough. Appreciated that you can take action quickly. I will finalise tomorrow with what we agreed above.

src/particle/particle/particle.py
eduardo-rodrigues: Added now.

PR #376, opened 2022-01-13
fix: use type annotations and TypeVar

henryiii:
This uncovered a few subclassabilty bugs.

I'd like to see if we can enable stricter type checking.


src/particle/particle/particle.py
```diff
@@ -262,16 +256,14 @@ def table_names(cls):
...
+    def table_loaded(cls) -> bool:
         """
         Check to see if the table is loaded.
         """
         return cls._table is not None
 
     @classmethod
-    def all(cls):
-        # type: () -> List[Particle]
+    def all(cls: Type[Self]) -> List[Self]:
```
eduardo-rodrigues, line 266: Just for my education: here you have `cls: Type[Self]` but in several class methods above the same is not necessary (in `table_loaded`, for example). Why is that?

src/particle/particle/particle.py
```diff
@@ -1049,7 +1017,7 @@ def from_name(cls, name):
             If no particle matches the input name.
         """
         try:
-            (particle,) = Particle.finditer(
+            (particle,) = cls.finditer(
```
eduardo-rodrigues, line 1020: Good catch!

src/particle/particle/utilities.py
```diff
@@ -35,8 +34,9 @@ def programmatic_name(name):
     return name.lstrip("_")
 
 
-def str_with_unc(value, upper, lower=None):
-    # type: (float, Optional[float], Optional[float]) -> str
+def str_with_unc(
+    value: float, upper: Optional[float], lower: Optional[float] = None
```
eduardo-rodrigues, line 38: Given the code should we not have ` upper: Optional[float] = None` instead?

eduardo-rodrigues (approved):
Looks great 👍! Thank you.

src/particle/particle/particle.py
henryiii: You don't need to annotate the first argument (self or cls); the reason we do it is to pin a TypeVar. TypeVar's _must_ be added to a "known" type before they can be used to describe a type. Since MyPy knows `cls`'s type, it can use that to inver the return type `List[Self]`. If `cls` is `Particle`, `Self` is an instance of Particle. If `cls` is a subclass of `Particle`, then `Self` is that subclass.

If [PEP 673](https://www.python.org/dev/peps/pep-0673/) is accepted, then we can do `from typing/typing_extensions import Self`, then drop `Self/Type[Self]` on self/cls and _just_ use it for returns, which would be _much_ nicer than all these TypeVar's. :) PEP 673 was submitted to the Python SC yesterday! :)

src/particle/particle/utilities.py
henryiii: We force users to be explicit, I think. `upper=None` means no uncertainty, while `lower=None` means upper and lower are the same. But maybe it's better with the default. Feel free to change if you want, it's just our utility, but not related to the PR. :)

eduardo-rodrigues: Indeed not. I will try and remember to look in the future. Depends on how we use the function. Fine as is, it seems.

PR #377, opened 2022-01-14
fix: full static typing

henryiii:
This finishes the typing and uses f-strings everywhere reasonable. Some `type: ignore`'s will go away after the next release of HEPUnits. Some minor polish continuing the removal of Python 2.

- fix(types): full static type checking
- style: switch remaining format strings to f-strings

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,10 +21,9 @@ profile = "black"
 [tool.mypy]
 warn_unused_configs = true
 warn_unused_ignores = true
-python_version = "3.6"
+python_version = "3.9"
 files = "src"
-check_untyped_defs = true
-disallow_untyped_defs = true
+strict = true
 
 # NumPy is used lightly, so ignored here for now
 [[tool.mypy.overrides]]
--- a/src/particle/particle/kinematics.py
+++ b/src/particle/particle/kinematics.py
@@ -8,9 +8,8 @@ Functions relevant to particle kinematics.
 """
 
 
-# Bug in HEPUnits exporting
-from hepunits.constants import hbar  # type: ignore[attr-defined]
-from hepunits.units import MeV, ns  # type: ignore[attr-defined]
+from hepunits.constants import hbar
+from hepunits.units import MeV, ns
 
 
 def width_to_lifetime(Gamma: float) -> float:
@@ -55,7 +54,7 @@ def width_to_lifetime(Gamma: float) -> float:
         return float("inf")
 
     # Just need to first make sure that the width is in the standard unit MeV
-    return hbar / float(Gamma / MeV)  # type: ignore[no-any-return]
+    return hbar / float(Gamma / MeV)
 
 
 def lifetime_to_width(tau: float) -> float:
@@ -101,4 +100,4 @@ def lifetime_to_width(tau: float) -> float:
         return float("inf")
 
     # Just need to first make sure that the lifetime is in the standard unit ns
-    return hbar / float(tau / ns)  # type: ignore[no-any-return]
+    return hbar / float(tau / ns)
--- a/src/particle/particle/kinematics.py
+++ b/src/particle/particle/kinematics.py
@@ -8,8 +8,9 @@ Functions relevant to particle kinematics.
 """
 
 
-from hepunits.constants import hbar
-from hepunits.units import MeV, ns
+# Bug in HEPUnits exporting
+from hepunits.constants import hbar  # type: ignore[attr-defined]
+from hepunits.units import MeV, ns  # type: ignore[attr-defined]
 
 
 def width_to_lifetime(Gamma: float) -> float:
@@ -54,7 +55,7 @@ def width_to_lifetime(Gamma: float) -> float:
         return float("inf")
 
     # Just need to first make sure that the width is in the standard unit MeV
-    return hbar / float(Gamma / MeV)
+    return hbar / float(Gamma / MeV)  # type: ignore[no-any-return]
 
 
 def lifetime_to_width(tau: float) -> float:
@@ -100,4 +101,4 @@ def lifetime_to_width(tau: float) -> float:
         return float("inf")
 
     # Just need to first make sure that the lifetime is in the standard unit ns
-    return hbar / float(tau / ns)
+    return hbar / float(tau / ns)  # type: ignore[no-any-return]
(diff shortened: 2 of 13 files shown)


pyproject.toml
```diff
@@ -21,10 +21,9 @@ profile = "black"
 [tool.mypy]
 warn_unused_configs = true
 warn_unused_ignores = true
-python_version = "3.6"
+python_version = "3.9"
```
henryiii, line 24: Targeting 3.9 gives better typing for `importlib.resources` (instead of `importlib_resources`), which is why I changed this here.

src/particle/particle/kinematics.py
```diff
@@ -8,8 +8,9 @@
 """
 
 
-from hepunits.constants import hbar
-from hepunits.units import MeV, ns
+# Bug in HEPUnits exporting
+from hepunits.constants import hbar  # type: ignore[attr-defined]
+from hepunits.units import MeV, ns  # type: ignore[attr-defined]
```
henryiii, line 13: These type ignores would go away with a new HEPUnits

src/particle/particle/kinematics.py
```diff
@@ -54,7 +55,7 @@ def width_to_lifetime(Gamma: float) -> float:
         return float("inf")
 
     # Just need to first make sure that the width is in the standard unit MeV
-    return hbar / float(Gamma / MeV)
+    return hbar / float(Gamma / MeV)  # type: ignore[no-any-return]
```
henryiii, line 58: Same for these, new HEPUnits would fix.

src/particle/particle/particle.py
```diff
@@ -1022,9 +1011,7 @@ def from_name(cls: Type[Self], name: str) -> Self:
             )  # throws an error if < 1 or > 1 particle is found
             return particle
         except ValueError:
-            raise ParticleNotFound(  # noqa: B904  <- use from None when Python 2 is dropped
-                f'Could not find name "{name}"'
-            )
+            raise ParticleNotFound(f'Could not find name "{name}"') from None
```
henryiii, line 1014: Better error messages (less unneeded printout)

eduardo-rodrigues: Yep. I confess _I have not yet browsed the code looking for comments such as this one related to Python 2. Good that you looked already 👍.

henryiii:
Okay, should be ready now, CI is green.

src/particle/particle/kinematics.py
eduardo-rodrigues: In that case I can go ahead and immediately tag and release `hepunits` ... Off I go and will cc you ...

henryiii: Okay, I'll rebuild and remove the type ignores as soon as that goes out.

eduardo-rodrigues: `hepunits` 2. 3.0 is out - https://pypi.org/project/hepunits/2.2.0/

src/particle/particle/particle.py
```diff
@@ -656,7 +654,7 @@ def L(self) -> Optional[int]:
         Note that the returned value corresponds to that effectively encoded
         in the particle PDG ID.
         """
-        return self.pdgid.L
+        return self.pdgid.L  # type: ignore[no-any-return]
```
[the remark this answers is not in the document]
eduardo-rodrigues: The issue here is that None is returned if somebody asks for L when it does not make sense. Return can be an int or None.

pyproject.toml
eduardo-rodrigues: Fair enough. I've also seen the odd place where we still use old-ish Python versions in checks. Will try and change as I "navigate around" ...

src/particle/particle/particle.py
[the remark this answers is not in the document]
henryiii: That's fine, the problem is self.pdgid is a PDGID instance, and PDGID's have all there methods filled in dynamically, at runtime, so you can't type check it. I'd rather use a solution that doesn't require a dynamic runtime cost, it's not something that changes that often. In fact, I'm not sure the standalone functions are needed anymore in Python 3 - Python 3 allows methods to be called on non-class instances.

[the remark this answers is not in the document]
henryiii: The return type is `Optional[int]`, but mypy has no idea what PDGID.L is, since it's dynamically added.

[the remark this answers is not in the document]
eduardo-rodrigues: Thanks for the discussion. Got it.

src/particle/particle/kinematics.py
```diff
@@ -8,6 +8,7 @@
 """
 
 
+# Bug in HEPUnits exporting
```
eduardo-rodrigues, line 11: You can remove this line.

pyproject.toml
henryiii: I usually run on the latest Python but target the oldest Python supported for checks. For high profile projects, sometimes we run mypy twice, once with oldest targeted, and once with newest targeted. Normally, oldest is fine, but if there's a reason checking is harder on the oldest (like here), I don't think there's a huge loss - we still have unit tests for older versions, we just won't get mypy immediately telling us if we use something from a newer Python.

src/particle/particle/kinematics.py
henryiii: Good catch, thanks!

src/particle/particle/particle.py
```diff
@@ -1269,7 +1255,7 @@ def _from_group_dict_list(cls: Type[Self], mat: Dict[str, Any]) -> List[Self]:
         if mat["state"] is not None:
             kw["J"] = float(mat["state"])
 
-        maxname = name + "({mat[mass]})".format(mat=mat) if mat["mass"] else name
+        maxname = name + f"({mat['mass']})" if mat["mass"] else name
```
eduardo-rodrigues, line 1258:
```suggestion
        maxname = name + f"({mat["mass"]})" if mat["mass"] else name
```

eduardo-rodrigues: Content-free suggestion. Purely for consistency.

eduardo-rodrigues (approved):
Really nice 👍!

src/particle/particle/particle.py
henryiii: The one directly above uses single on the outside and double in the middle.

eduardo-rodrigues: Oh boy, I was far too quick. Of course. Think so too for outer string single quoted.

PR #382, opened 2022-02-11
chore: clean up config

henryiii:
I didn't realize .cfg files get interpolated, but they do (see https://github.com/twisted/twisted/blob/trunk/setup.cfg).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,30 +51,20 @@ package_dir =
 where = src
 
 [options.extras_require]
-all =
-    check-manifest>=0.42
-    pandas
-    pytest>=6
-    pytest-benchmark
-    tabulate
-dev =
-    check-manifest>=0.42
-    pandas
-    pytest>=6
-    pytest-benchmark
-    tabulate
 test =
     pandas
     pytest>=6
     pytest-benchmark
     pytest-cov
     tabulate
+dev =
+    check-manifest>=0.42
+    %(test)s
+all =
+    $(dev)s
 
 [options.package_data]
 * = *.csv, *.fwf, *.mcd, *.py.typed
-
-[bdist_wheel]
-universal = 1
 
 [tool:pytest]
 testpaths =


henryiii:
Hmm, going to have to think about setup-cfg-fmt here, it doesn't handle interpolation. I can just leave this one, but thinking about using it elsewhere too. Hmmm....

eduardo-rodrigues:
Hey @henryiii, I will get back to you on the various fronts in Particle early next week. I have to say I do not know what you mean by ".cfg files get interpolated".

setup.cfg
```diff
@@ -38,9 +38,13 @@ keywords =
 packages = find:
 install_requires =
     attrs>=19.2
-    deprecated
     hepunits>=2.0.0
+    check-manifest;extra == dev" or extra == "all"
     importlib-resources>=2.0;python_version<"3.9"
+    pytest>=6;extra == "test" or extra == dev" or extra == "all"
```
abravalheri, line 44: Maybe it is missing a `"` before `dev`? (In all items that seem to be the case)

abravalheri: (Although, it might still not work, because the way setuptools process this is... [complicated](https://github.com/pypa/setuptools/blob/19bfd4d13755011a027013a63a637f02bda0c0cd/setuptools/dist.py#L528)...)

henryiii: I think that was it, seems to be passing now!

henryiii:
I rebased it. This is more an experiment to see if this is valid for our other repos - which it is, but it's also kind of fighting setup-cfg-fmt. It would be a useful idea if we / when we move to PEP 621 config. I think I'm going to go a slightly different path, though, and instead remove the suggestion that we provide "dev" and "all" extras, and try to keep the extras as separate and non-overlapping, as it just simply is always going to be a fight with Python to make combined extras, and combined environments aren't really that useful (better to keep separate environments via nox or pre-commit).

abravalheri:
@henryiii we could write a PEP for that at some point in the future... The suggestion of the `{include = ...}` syntax in `pyproject.toml project.optional-dependencies` suggested in the discord is really good.

PR #387, opened 2022-03-28
MANIFEST.in: Include tests in sdist

### LINKED ISSUE #386: Include the tests in the pypi tarball please
I'd like to add this package to NixOS, and I noticed you don't include the tests in the pypi tarball. I also noticed that I can't use the GitHub tarball since it doesn't contain a `version.py` file, and also `setuptools-scm` can't detect the version when I use a GitHub tarball. NixOS also has an issue with using git checkouts (see https://github.com/NixOS/nixpkgs/pull/4752) that would have helped `setuptools-scm`.

Thanks in advance.

doronbehar:
Is it as simple as that? cc @henryiii .

Closes #386.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -20,3 +20,5 @@ include src/particle/version.py
 include LICENSE README.rst CONTRIBUTING.md pyproject.toml setup.py setup.cfg .zenodo.json
 
 recursive-include src py.typed
+
+recursive-include tests *.py *.mcd *.rst


eduardo-rodrigues:
Henry can confirm but I think you meant to write `recursive-include tests *.py *.mcd *.rst`, the 2 latter to ensure that what is under `tests/data` gets added as well.

doronbehar:
> I think you meant to write

Done.

eduardo-rodrigues (approved):

henryiii (approved):

PR #389, opened 2022-03-28
chore: minor cleanup and updates

henryiii:
This captures the good parts of #382, without trying to be more clever with config.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -62,3 +62,14 @@ repos:
   hooks:
   - id: codespell
     exclude: ^(src/particle/data/nuclei2020.csv|src/particle/data/mass_width_2008.fwf)$
+
+- repo: https://github.com/pre-commit/pygrep-hooks
+  rev: v1.9.0
+  hooks:
+  - id: python-check-blanket-noqa
+  - id: python-check-blanket-type-ignore
+  - id: python-no-eval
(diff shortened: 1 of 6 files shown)


henryiii:
FYI, I'm using `scikit-hep-repo-review` (experimental side project) to evaluate the current status of compliance to the scikit-hep developer guidelines and have addressed the issues shown so far except for isort.

Screen Shot 2022-03-28 at 11 52 46 AM

henryiii:
Okay to go in?

henryiii:
We can always re-add "deprecated", but it was making it harder to install particle on WebAssembly, so I had dropped it since we are literally not currently using it. We can re-add it if we want to use it again, it's just two lines, but adding a dependency just for possible future use isn't really helpful. Removing the `universal=True` setting I would consider important, since it's causing us to produce `py2.py3` filenames for our releases.

eduardo-rodrigues (approved):

eduardo-rodrigues:
True. Fine, then move forward ...
BTW, you remove
```
[bdist_wheel]
universal = 1
```
but this package *is* a universal package. What's the catch I'm not aware of?

henryiii:
"Universal" package specifically means it runs on Python 2 and Python 3 without changes. That's not true, is it? That's why wheel adds python 2 tags to the wheel names when you set universal=1.

PR #399, opened 2022-05-09
dependabot.yml - run monthly rather than daily

eduardo-rodrigues:


henryiii:
These have only come up once every few months or so from looking at the history, so not sure we really gain anything by making it weekly?

eduardo-rodrigues:
> These have only come up once every few months or so from looking at the history, so not sure we really gain anything by making it weekly?

We certainly gain from going daily to weekly with this MR - daily was totally unnecessary ;-). But let's go monthly.

PR #420, opened 2022-10-04
feat: drop Python 3.6, add 3.11

henryiii:
- feat: drop 3.6, add 3.11
- style: run pre-commit
- tests: drop no-op use of path as context manager (deprecated in 3.11)


eduardo-rodrigues (approved):
Thank you for this!

Cool to see the new and simpler annotations.

PR #421, opened 2022-10-05
chore: cleanup with reverb suggestions

henryiii:
Some suggestions using `reverb`.


eduardo-rodrigues:
I'm not sure some of the changes elif to if are really better but the changes do not affect the outcome, so OK. Merge when happy.

PR #423, opened 2022-10-05
feat: move to hatchling

henryiii:
Moving over to using hatchling.


eduardo-rodrigues:
I have to go and will look into this later ... As a general comment - it would make sense to me to first discuss changes/moves such as this one among us prior to being exposed to the fact. I can't see a counter argument. Kindly have this in mind in the future.

henryiii:
Sorry, this has been happening across other Scikit-HEP repos, like https://github.com/scikit-hep/uproot5/pull/688, https://github.com/scikit-hep/hist/pull/418, and https://github.com/scikit-hep/vector/pull/223. Didn't know it would need further discussion here. I'm interested in moving all of our pure Python packages over.

henryiii:
And have no idea why Actions isn't running anything ATM.

pyproject.toml
```diff
@@ -1,23 +1,80 @@
...
-[tool.setuptools_scm]
-write_to = "src/particle/version.py"
+[project]
+name = "particle"
+description = "Extended PDG particle data and MC identification codes"
+readme = "README.rst"
+requires-python = ">=3.7"
+authors = [
+    { name = "Eduardo Rodrigues", email = "eduardo.rodrigues@cern.ch" },
+    { name = "Henry Schreiner", email = "hschrein@cern.ch" },
```
eduardo-rodrigues, line 12: Do you want to update your email address?

eduardo-rodrigues (approved):
I'm unfamiliar with hatchling but seems fine from reading changes themselves. It's kind of self-explanatory.

No more need for the MANIFEST file?

pyproject.toml
henryiii:
```suggestion
    { name = "Henry Schreiner", email = "henryfs@princeton.edu" },
```

Still valid but okay, maybe this is better.

henryiii:
That’s a setuptools thing. The SDist will have a bit more in it, but that’s fine. The wheel should be just like before. Hatchling uses the .gitignore file to decide what to include. You can manually add config to the pyproject file to exclude or add files, but it does a good job by default.

PR #424, opened 2022-10-08
chore: add pylint

henryiii:
Adding pylint.

### DIFF
--- a/src/particle/converters/bimap.py
+++ b/src/particle/converters/bimap.py
@@ -85,7 +85,7 @@ class BiMap(Generic[A, B]):
         elif isinstance(filename, HasOpen):
             file_object = filename.open()
         else:
-            file_object = open(filename)  # type: ignore[arg-type]
+            file_object = open(filename, encoding="utf_8")  # type: ignore[arg-type]
 
         with file_object as _f:
             self._to_map = {
@@ -175,7 +175,7 @@ def DirectionalMaps(
     elif isinstance(filename, HasRead):
         file_object = filename
     else:
-        file_object = open(filename)  # type: ignore[arg-type]
+        file_object = open(filename, encoding="utf_8")  # type: ignore[arg-type]
 
     with file_object as _f:
         skipinitialspace = True
@@ -211,6 +211,7 @@ else:
 
 
 class DirectionalMap(StrStrMapping):
+    # pylint: disable-next=redefined-builtin
     def __init__(self, name_A: str, name_B: str, map: dict[str, str]) -> None:
         """
         Directional map class providing a A -> B mapping.
--- a/src/particle/particle/convert.py
+++ b/src/particle/particle/convert.py
@@ -106,7 +106,8 @@ def filter_file(fileobject: StringOrIO) -> TextIO:
 
     if not hasattr(fileobject, "read"):
         assert isinstance(fileobject, str)
-        fileobject = open(fileobject, encoding="utf-8")
+        with open(fileobject, encoding="utf-8") as f:
+            return filter_file(f)
 
     assert not isinstance(fileobject, (str, Traversable))
 
@@ -116,9 +117,6 @@ def filter_file(fileobject: StringOrIO) -> TextIO:
         if not line.lstrip("\ufeff").lstrip().startswith("*"):
             stream.write(line)
     stream.seek(0)
-
-    if not fileobject.closed:
-        fileobject.close()
 
     return stream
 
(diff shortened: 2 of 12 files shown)


src/particle/particle/convert.py
```diff
@@ -106,7 +106,8 @@ def filter_file(fileobject: StringOrIO) -> TextIO:
 
     if not hasattr(fileobject, "read"):
         assert isinstance(fileobject, str)
-        fileobject = open(fileobject, encoding="utf-8")
+        with open(fileobject, encoding="utf-8") as f:
+            return filter_file(f)
```
henryiii, line 110: This handles the file by opening & _always_ closing it only if it's not already open. Otherwise it leaves it alone.

src/particle/converters/bimap.py
```diff
@@ -85,7 +85,7 @@ def __init__(
         elif isinstance(filename, HasOpen):
             file_object = filename.open()
         else:
-            file_object = open(filename)  # type: ignore[arg-type]
+            file_object = open(filename, encoding="utf_8")  # type: ignore[arg-type]
```
henryiii, line 88: Default encoding should always be specified, will be changed to be always unicode in Python 3.15.

src/particle/particle/convert.py
eduardo-rodrigues: Good point.

src/particle/converters/bimap.py
eduardo-rodrigues: We're good for a while but being explicit is hardly ever bad 👍.

eduardo-rodrigues (approved):
I still do not understand the preference for a series of if's and never use elif, but fine, the functions remain correct with the changes.

Thanks for this addition of pylint.

henryiii:
It's not "never use elif", just don't add useless logic complication if it's not needed. You still need elif/else if you don't have a control break. There's no need for an "else" after a control breaking statement (`return`, `raise`, `break`, `continue`). It's just extra text and indentation that is unnecessary. While it is a bit harder to see in the diff, it is much less mental work (and correspondingly much simpler AST) when you don't chain these on, and instead think of them as guards. It also highlights the "fall through" case by not indenting it.

Here's an example to show how it scales:

```python
def calcPercentage(number):
    if isinstance(number, float):
        if number > 1:
            return "Number must be less than 1"
        else:
            return f"{number:%}"
    else:
        return "Please enter valid whole number"
```

Notice how you have to check every possible branch to see if everything is covered. You have a "dangling else" at the end that just floats there - in a long function, you might not be able to even see what it lines up with without scrolling.

Now compare with just removing the else's:

```python
def calcPercentage(number):
    if isinstance(number, float):
        if number > 1:
            return "Number must be less than 1"
        return f"{number:%}"
    return "Please enter valid whole number"
```

Without looking at the logic, you can visually see that this function will never "fall through" and accidentally return None. All branches are clearly accounted for. Dangling blocks are removed. However, this is still not "guard pattern". Let's rewrite again:

```python
def calcPercentage(number):
    if not isinstance(number, float):
        return "Please enter valid whole number"

    if number > 1:
        return "Number must be less than 1"

    return f"{number:%}"
```

Now we've lost a whole level of indention, and both statements logically "guard" the flow to the answer. Mentally, you only have to consider one guard at a time - while elseing them together would require you again view the whole block. This is more modular, one action (a guard check) at a time.

It think the above is ideal in this case, but if you really felt the last two items were similar and wanted to show that, I'd write it this way:

```python
def calcPercentage(number):
    if not isinstance(number, float):
        return "Please enter valid whole number"

    return f"{number:%}" if number <= 1 else "Number must be less than 1"
```

Which still avoids adding an "else" when it's meaningless.

henryiii:
FYI, Ruby has the guard pattern built in as syntax, it would look like this:

```ruby
def calcPercentage(number)
    return "Please enter valid whole number" unless number.is_a? Float
    return "Number must be less than 1" if number > 1

    return "#{number * 100}%"
end
```

henryiii:
And here's an example from this PR - I really think the diffs make it hard to see the benefit:

Before:
```python
    def _str_mass(self) -> str:
        if self.mass is None:
            return "None"
        else:
            txt = str_with_unc(self.mass, self.mass_upper, self.mass_lower)
            return f"{txt} MeV"
```

After:
```python
    def _str_mass(self) -> str:
        if self.mass is None:
            return "None"

        txt = str_with_unc(self.mass, self.mass_upper, self.mass_lower)
        return f"{txt} MeV"
```

eduardo-rodrigues:
Thank you Henry for taking the time to give some good examples. Points taken and indeed the rewritten versions are smoother and clear to read 👍.

eduardo-rodrigues (approved):

PR #445, opened 2022-12-15
Finalise coverage of particle.utilities

eduardo-rodrigues:


eduardo-rodrigues:
Wrong approach. Will start clean again.

PR #448, opened 2022-12-15
Full test of Particle.from_nucleus_info() method

eduardo-rodrigues:
FYI @The-Ludwig I'm here improving the tests for the `Particle.from_nucleus_info()` method you enhanced the package with. We now have full coverage of the method.


The-Ludwig:
Happy to see my contributions are still used! :)

PR #462, opened 2023-01-17
Wrong corsika7id to PDGid converting

The-Ludwig:
I noticed that the corsika ids of all `Sigma(c)`s and `Sigma(c)star`s got mapped to the same pdgid, since pdg does not name the sigma(c)s with stars, rather with the masses of the two types of sigma(c)s. I changed it setting corsika `Sigma(c)` as pdgid `Sigma(c)(2455)` and `Sigma(c)*` as `Sigma(c)(2520)`.


The-Ludwig:
Edit: id for omega_c was also wrong...

The-Ludwig:
pre-commit failing on files I did not change?

henryiii:
This will be fixed by the next release of flake8-bugbear. They added a new default check that's quickly being moved to an optional check.

The-Ludwig:
I also deleted a weird empty test file which I apperantly created. This is ready to merge, from my point of view. Changes are quite simple...

eduardo-rodrigues:
Thanks a lot @The-Ludwig. I will review this week.

eduardo-rodrigues:
Hi @The-Ludwig, that's right, see for example https://github.com/scikit-hep/particle/blob/master/src/particle/data/conversions.csv#L293-L312. The fact that we did not notice the duplication also means there was no test to verify no duplications sneak in. Maybe worth adding for future robustness anyway?

eduardo-rodrigues:
[If you rebase the pre-commit error will be gone as I fixed that on master.]

The-Ludwig:
@eduardo-rodrigues Yes, you're right I also thought about it. Then I thought some more about how to test for errors in the table and the main reason why this error slipped in is, because no one double checked my hand transcribed translation table. If there happens to be some translation error in the table which is not a duplicate, we won't catch it. 
Btw: checking the table is straight forward, since I used clear names, it's possible to open the Corsica User Guide next to the table and look for errors. 

Of course no reason not to implement more tests!

admin/dump_pdgid_to_corsika7.py
```diff
@@ -92,18 +92,18 @@
     (157, common_particles["Omega_c_0_bar"]),
     (119, common_particles["D_0_bar"]),
     (120, common_particles["D_s_plus"]),
-    (161, common_particles["Sigma_c_2455_pp"]),
+    (161, common_particles["Sigma_c_2520_pp"]),
```
eduardo-rodrigues, line 95: Are you sure of this and similar changes? As I know nothing about Corsika, where can these mappings be checked?

I'm doubting because I then see no matching for the lowest level c-baryons and only for excited ones. So (see https://pdglive.lbl.gov/ParticleGroup.action?init=0&node=BXXX040) the Sigma_c ground level is identified by the mass 2455, the excited Sigma_c^* one with 2520, and the Xi_c^* with 2645.

I let you comment and check it all as you have all info in hands :). Let me know. Thanks a lot.

The-Ludwig: You can check the Corsika7IDs in the wonderful and very comprehensive Corsika userguide on page 120 and 121 (pdf page 132 and 133) https://web.iap.kit.edu/corsika/usersguide/usersguide.pdf

The Sigma_c (2455) states are in the script with IDs 140-142 and 152-154. (Eventually I should print the list in an ordered version... 😅)

Thanks for going throught the PR and helping me improve (Particle as well as my own coding style)! :)

eduardo-rodrigues:
> @eduardo-rodrigues Yes, you're right I also thought about it. Then I thought some more about how to test for errors in the table and the main reason why this error slipped in is, because no one double checked my hand transcribed translation table. If there happens to be some translation error in the table which is not a duplicate, we won't catch it. Btw: checking the table is straight forward, since I used clear names, it's possible to open the Corsica User Guide next to the table and look for errors.
> 
> Of course no reason not to implement more tests!

I see. I take it also on me as I did not check the long list of numbers in the first commit. It was long and ... Point taken.

eduardo-rodrigues (approved):
Great, I rechecked with the userguide, being maniac.

We can follow-up on this to have the pairs in order and also add in the header docstring the link to the user guide, for reference.

Merging now.

PR #469, opened 2023-02-06
chore: clean up zipfile

henryiii:
This was generating .pyc files for all packages; those are not needed (since they will target only one version of Python - 3.7 in this case).

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -57,11 +57,10 @@ def zipapp(session: nox.Session) -> None:
         "-m",
         "pip",
         "install",
+        "--no-compile",
         ".",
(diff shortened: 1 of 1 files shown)


eduardo-rodrigues (approved):
Sure you can remove "attrs" and "hepunits"? I trust you, go ahead.

henryiii:
Anything that's a dependency is fine, it automatically gets installed into the target directory when you install the package. I list the conditional dependencies (typing_extensions & importlib_metdata) because they won't get automatically installed if we don't use Python 3.7 (which we do, in the nox job, but just in case).

Issue #476, opened 2023-03-03
Confusing `is_quarkonium` in Readme


APN-Pucky:
From wikipedia: Quarkonium is a [flavorless](https://en.wikipedia.org/wiki/Flavor_(physics)) [meson](https://en.wikipedia.org/wiki/Meson) whose constituents are a heavy [quark](https://en.wikipedia.org/wiki/Quark) and its own [antiquark](https://en.wikipedia.org/wiki/Antiquark) [1].

But lambdas from the readme are:
```
>>> from particle.pdgid import is_meson, three_charge, has_charm, has_bottom, has_top
>>> is_heavy_flavor = lambda x: has_charm(x) or has_bottom(x) or has_top(x)
>>> is_quarkonium = lambda x: is_meson(x) and three_charge(x)==0 and is_heavy_flavor(x)
>>> is_quarkonium(421)
True
```
Do you have a different definition of a quarkonium? since:
```
>>> Particle.from_pdgid(421)
<Particle: name="D0", pdgid=421, mass=1864.84 ± 0.05 MeV>
>>> p=_
>>> p.quarks
'cU'
```

[1] https://en.wikipedia.org/wiki/Quarkonium

APN-Pucky:
I would rather consider these quarkonia:
```
>>> f=Particle.finditer(lambda p: len(p.quarks)==2 and p.quarks[0] == p.quarks[1].swapcase())
>>> for p in f:
...     print(p.pdgid)
... 
<PDGID: 441>
<PDGID: 443>
<PDGID: 445>
<PDGID: 553>
<PDGID: 555>
<PDGID: 10441>
<PDGID: 10443>
<PDGID: 10551>
<PDGID: 10553>
<PDGID: 20443>
<PDGID: 20553>
<PDGID: 20555>
<PDGID: 30443>
<PDGID: 100441>
<PDGID: 100443>
<PDGID: 100445>
<PDGID: 100553>
<PDGID: 100555>
<PDGID: 110551>
<PDGID: 120553>
<PDGID: 200553>
<PDGID: 300553>
<PDGID: 9000443>
<PDGID: 9000553>
<PDGID: 9010443>
<PDGID: 9010553>
```

eduardo-rodrigues:
Thank you @APN-Pucky for reporting. Seems indeed fishy. It could be that we introduced a couple of hints in code that are not totally to the point. I will look into this next week ...

[...]

PR #477, opened 2023-03-03
Fix is_quarkonium in Readme

APN-Pucky:
Closes: https://github.com/scikit-hep/particle/issues/476

This fixes the example such that we only get quarkonia 
```
>>> print ([p.quarks for p in Particle.finditer(lambda p: len(p.quarks)==2 and p.quarks[0] == p.quarks[1].swapcase())])
['cC', 'cC', 'cC', 'bB', 'bB', 'cC', 'cC', 'bB', 'bB', 'cC', 'bB', 'bB', 'cC', 'cC', 'cC', 'cC', 'bB', 'bB', 'bB', 'bB', 'bB', 'bB', 'cC', 'bB', 'cC', 'bB', 'cC']
>>> print ([p.quarks for p in Particle.finditer(lambda p: is_meson(p.pdgid) and is_heavy_flavor(p.pdgid) and p.is_self_conjugate)])
['cC', 'cC', 'cC', 'bB', 'bB', 'cC', 'cC', 'bB', 'bB', 'cC', 'bB', 'bB', 'cC', 'cC', 'cC', 'cC', 'bB', 'bB', 'bB', 'bB', 'bB', 'bB', 'cC', 'bB', 'cC', 'bB', 'cC']
```
instead of what current example produces
```
>>> print ([p.quarks for p in Particle.finditer(lambda p: is_meson(p.pdgid) and is_heavy_flavor(p.pdgid) and three_charge(p.pdgid)==0)])
['cU', 'Cu', 'cU', 'Cu', 'cU', 'Cu', 'cC', 'cC', 'cC', 'dB', 'Db', 'dB', 'Db', 'dB', 'Db', 'sB', 'Sb', 'sB', 'Sb', 'sB', 'Sb', 'bB', 'bB', 'cU', 'Cu', 'cU', 'Cu', 'cC', 'cC', 'bB', 'bB', 'cU', 'Cu', 'cC', 'bB', 'bB', 'cC', 'cC', 'cC', 'cC', 'bB', 'bB', 'bB', 'bB', 'bB', 'bB', 'cC', 'bB', 'cC', 'bB', 'cC']

```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -83,7 +83,7 @@ For the sake of example, quarkonia can be specified with the following user-defi
 .. code-block:: python
 
     >>> is_heavy_flavor = lambda x: has_charm(x) or has_bottom(x) or has_top(x)
-    >>> is_quarkonium = lambda x: is_meson(x) and three_charge(x)==0 and is_heavy_flavor(x).
+    >>> is_quarkonium = lambda x: is_meson(x) and is_heavy_flavor(x) and Particle.from_pdgid(x).is_self_conjugate
 
 PDG ID literals provide (``PDGID`` class) aliases for all particles loaded, with easily recognisable names.
 For example:


eduardo-rodrigues (approved):
Thanks again for spotting the mistake!

I'm going to merge this as-is though the solution is not perfect in the place in the docs it's in, because it says "These composable functions qualifying PDG IDs make it easy to classify particles. For the sake of example, quarkonia can be specified with the following user-defined functions:" whereas this particular case in fact requires the self-conjugation information that requires a Particle instance, not just the PDG ID. It's an example that should rather be moved to a discussion on the Particle class. I will think this better for the future ...

PR #494, opened 2023-05-09
Fix j_spin for susy particle

### LINKED ISSUE #463: spin of SUSY partners
While checking out the package, I found that some of the SUSY partners of the SM quarks are assigned a spin value of 2 (that in the 2s+1 scheme represents a spin-1/2 particle). For instance, the left-handed partner of up-quark `suL`, which is usually represented by id 1000002, is assigned a value of 2 for its `j_spin`. 
But the squarks are bosons, so their spins are integral and should be 1 or 3.

Am I missing something? Please let me know if this is a known (and intentional) issue.

amanmdesai:
This PR aims to fix #463 

@eduardo-rodrigues 
This is a proposed solution. There is some work to be done for example to check if the spin of all sfermions is indeed zero. As of now PR only fixes the issue with sfermions and not with gluinos. 
 Please let me know if the strategy is okay  

Thanks

### DIFF
--- a/src/particle/pdgid/functions.py
+++ b/src/particle/pdgid/functions.py
@@ -710,6 +710,8 @@ def j_spin(pdgid: PDGID_TYPE) -> int | None:
                 return 2
             if 22 <= fund < 38:
                 return 2
+            if fund == 39:
+                return 4
         else:  # other particles
             if 0 < fund < 7:  # 4th generation quarks not dealt with !
                 return 2
--- a/src/particle/pdgid/functions.py
+++ b/src/particle/pdgid/functions.py
@@ -705,7 +705,7 @@ def j_spin(pdgid: PDGID_TYPE) -> int | None:
         fund = _fundamental_id(pdgid)
         if is_SUSY(pdgid):  # susy particles
             if 0 < fund < 17:
-                return 0
+                return 1
             if fund == 21:
                 return 2
             if 22 <= fund < 38:
--- a/src/particle/pdgid/functions.py
+++ b/src/particle/pdgid/functions.py
@@ -703,17 +703,21 @@ def j_spin(pdgid: PDGID_TYPE) -> int | None:
         return None
     if _fundamental_id(pdgid) > 0:
         fund = _fundamental_id(pdgid)
-        if 0 < fund < 7:  # 4th generation quarks not dealt with !
-            return 2
-        if (
-            fund == 9
-        ):  # Alternative ID for the gluon in codes for glueballs to allow a notation in close analogy with that of hadrons
-            return 3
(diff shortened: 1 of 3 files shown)


eduardo-rodrigues:
Hello. That was quick! I'm low on bandwidth this week so let me just say that this needs to be complemented with dedicated tests. Some tests are around in several places but there is some enhancement required. Let me know if you do not see the places where relevant tests are or are to be added ...

Many thanks. Back asap.

amanmdesai:
Hello @eduardo-rodrigues Thanks for your message.
Sure I will explore the possible tests that can be added related to susy particles.

amanmdesai:
I am now modifying the existing tests.
It would be nice and helpful if you could approve the workflows.

Thanks,
Aman

amanmdesai:
All particles listed in the susy table in PDG (monte carlo numbering note) are covered with the exception of PDG ID 1000039. @eduardo-rodrigues What is spin for this particle (Particle ID 1000039)? 

The tests are modified to include a few cases.

amanmdesai:
Hi @eduardo-rodrigues This PR is now ready for review

eduardo-rodrigues:
Thank you very much @amanmdesai. I'm rather loaded but we will finalise this by the end of the week.

amanmdesai:
> Thank you very much @amanmdesai. I'm rather loaded but we will finalise this by the end of the week.

Thanks @eduardo-rodrigues

eduardo-rodrigues (approved):
This all looks correct and great to me. Many thanks, again!

It would be interesting to check SUSY particles better in the package, e.g. for lepton and baryon number, but that's a separate matter.

amanmdesai:
> This all looks correct and great to me. Many thanks, again!

Thanks!

> It would be interesting to check SUSY particles better in the package, e.g. for lepton and baryon number, but that's a separate matter.

@eduardo-rodrigues This sounds interesting and will be useful as well. Shall I implement these tests in a separate PR? Could you please guide me on how lepton numbers are extracted using this package? 

Also would it be useful to introduce a function (and relevant tests) to extract the R partity of a particle?

Thanks,
Aman

amanmdesai:
> It would certainly be interesting to expand on certain aspects in the package. Some extensions are not trivial and require care. Maybe better have tasks for these?



> Let me ping @redeboer whom I believe has been using lepton and baryon numbers in his work around the ComPWA project. It would actually be interesting to understand what Particle can provide as ready-to-use info from PDG IDs, in particular, to avoid duplications. Again, better discuss such matters in tasks :).
> 
> +1

Sure we can discuss this in a new issue?

eduardo-rodrigues:
Indeed best.

PR #508, opened 2023-07-03
chore: ruff moved to astral-sh

henryiii:
See https://github.com/scientific-python/cookie/pull/205.

Committed via https://github.com/asottile/all-repos


eduardo-rodrigues (approved):
Thanks.

PR #536, opened 2023-11-16
chore: modernize Ruff config, use ruff-format

henryiii:
Modernize Ruff config. ~~Waiting on the next release (which adds better Jupyter notebook support) for ruff-format.~~ (Now released)

See https://github.com/scientific-python/cookie/pull/299.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,9 @@ on:
     branches:
     - master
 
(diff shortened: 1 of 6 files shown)


henryiii:
Does anything need to be done here?

eduardo-rodrigues (approved):
Hi. All good to me. I simply forgot to reply.

PR #581, opened 2024-04-01
chore: require Python 3.8+

henryiii:
- **chore: drop Python 3.7, add Python 3.12**

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -38,7 +38,7 @@ jobs:
 
     - uses: wntrblm/nox@2024.03.02
       with:
-        python-versions: "3.7"
+        python-versions: "3.8"
 
     - name: Make ZipApp
       run: nox -s zipapp
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,16 +52,11 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: Setup uv
-      uses: yezz123/setup-uv@v4
-      with:
-        uv-venv: ".venv"
-
     - name: Install package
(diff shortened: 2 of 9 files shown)


henryiii:
Time for most jobs is roughly cut in half, from about 30 seconds to 15 seconds.

eduardo-rodrigues (changes requested):
It is time indeed to remove 3.7 support though I had not done it since "no real need". Fair enough at this stage.

I don't really like to have PRs mixing 2 things such as here. Also, I do not like the idea to move to a super new thing that is basically a 1-person project when there is no problem with installing the standard and 100% supported way. I was OK with moving to Ruff as it got huge support, but uv is super new. There simply is no real gain here, just a danger.

henryiii:
Sure, there were two unrelated commits, I just revered the uv one.

If you are talking about uv, it's not a "1-person project"; it has between 5-7 full time developers. It has more GitHub stars than pip. It's had great adoption, including by Jupyter. Nox, tox, Pixi, and build all support it. build-and-inspect-python-package now uses it.

It's also just in CI; if there's an issue, we can remove the letters "uv" and keep going - I'd say there's very little "danger". It cuts the time waiting for a CI result in half, so I wouldn't say "no gain".

If you are referring to `yezz123/setup-uv`, yes, as far as I know, that's a one person project. But it's just downloading uv and optionally setting up an environment, so it's easy to replace if it breaks.

eduardo-rodrigues:
I was indeed talking about yezz123/setup-uv. I see, thanks for the info. Let's then go with uv in a future PR. (Can that be done without yezz123/setup-uv? Otherwise, fine, I trust you.)

eduardo-rodrigues (approved):
Feel free to merge when happy.

PR #590, opened 2024-04-23
Prepare for release 0.24.0

eduardo-rodrigues:


eduardo-rodrigues:
Hello @henryiii, I see that this new release did not get deployed to PyPI. I also see that for example cd.yml in Hist got some updates. Can you help out?

henryiii:
I downloaded the artifact and manually uploaded it via a temporary token. I've also entered this repo and `cd.yml` into the trusted publisher setup on PyPI. I've specified a not-currently-existing `pypi` environment as the release environment - it will be created on the first release with it. Now I'll make the changes from https://learn.scientific-python.org/development/guides/gha-pure/.

eduardo-rodrigues:
Brilliant, thank you so much!

PR #591, opened 2024-04-23
ci: use PyPI trusted publisher

henryiii:
See https://github.com/scikit-hep/particle/pull/590.


eduardo-rodrigues (approved):
Thank you. Merge when you see fit.

PR #592, opened 2024-04-24
ci: macos-latest is changing to macos-14 ARM runners

henryiii:
Committed via https://github.com/asottile/all-repos


eduardo-rodrigues (approved):
Fair enough, thank you.

PR #613, opened 2024-08-09
ci: fix release file upload

### LINKED ISSUE #612: ZipApp is not being published since version 0.24.0
Should it be simply removed at this stage? I don't think it's being used by anyone (at least that I'm aware of).

henryiii:
Should fix #612, I think.


eduardo-rodrigues (approved):
Thanks🤞!

PR #631, opened 2024-09-29
ci: add attestations

henryiii:
The PyPI ones are not persisted yet, but they are checked.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -53,10 +53,12 @@ jobs:
     needs: [dist]
     runs-on: ubuntu-latest
     if: github.event_name == 'release' && github.event.action == 'published'
-    environment: pypi
+    environment:
+      name: pypi
+      url: https://pypi.org/p/particle
     permissions:
       id-token: write
-
+      attestations: write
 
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -53,18 +53,28 @@ jobs:
...
 
     steps:
     - uses: actions/download-artifact@v4
       with:
         name: artifact
         path: dist
 
-    - uses: pypa/gh-action-pypi-publish@release/v1
+    - name: Generate artifact attestation for sdist and wheel
+      uses: actions/attest-build-provenance@1c608d11d69870c2092266b3f9a6f3abbf17002c # v1.4.3
```
eduardo-rodrigues, line 70: I hope all this commit hashes will not become a pain to update?

henryiii: Dependabot should do it. Though we can use any other form, but this is common for this action since it's security related.

eduardo-rodrigues: OK.

eduardo-rodrigues (approved):

Issue #672, opened 2025-05-10
Higgs J spin is None


APN-Pucky:
```py
>>> print(particle.Particle.from_pdgid(21).J)
1.0
>>> print(particle.Particle.from_pdgid(22).J)
1.0
>>> print(particle.Particle.from_pdgid(24).J)
1.0
>>> print(particle.Particle.from_pdgid(25).J)
None
```

Shouldn't this return 0 instead of None?

eduardo-rodrigues:
Hi @APN-Pucky, yes, well spotted! The "funny" thing is that the spin is correct in the data files. So indeed the logic of the J method of the PDGID class was forgetting the case of the Higgs! I see you already committed the fix. Thanks!

APN-Pucky:
> The "funny" thing is that the spin is correct in the data files.

What's the point of having both? I guess SUSY is not in the data files. If a user were to supply custom data files shouldn't the data files take precedence?

PR #673, opened 2025-05-10
fix: higgs j_spin is zero

APN-Pucky:
Closes: #672


eduardo-rodrigues (approved):
Hi @APN-Pucky, yes, well spotted! The "funny" thing is that the spin is correct in the data files. So indeed the logic of the J method of the PDGID class was forgetting the case of the Higgs! I see you already committed the fix. Thanks!

eduardo-rodrigues:
Let me know if you need a release urgently. Otherwise I will release later in the week, tomorrow or so.

eduardo-rodrigues:
@all-contributors please add @APN-Pucky for code.

Issue #671, opened 2025-05-08
Possibly a better way to isolate digits of a pdg id?


TanmayPani:
While trying to efficiently determine charges of a large awkward array of pdg id codes, I wanted to use this library in a numba jit function, but it seems like the lines 902, 903 in the `_digit` function in `particle/src/particle/pdgid/functions.py`:
```
def _digit(pdgid: PDGID_TYPE, loc: int) -> int:
    """
    Provides a convenient index into the PDGID number, whose format is in base 10.

    Returns the digit at position 'loc' given that the right-most digit is at position 1.
    """
    sid = str(abspid(pdgid))
    return int(sid[-loc]) if loc <= len(sid) else 0
```

Are preventing jitting as numba doesnt support direct conversion from `string/char` to `int`, but as even the workaround of conversion from string to bytearray to int is unnecessary overhead, is there a reason why the digits are being isolated in this manner rather than something like:
```
sid = (abspid(pdgid) % 10**loc) // (10 ** (loc - 1))
return nb.int32(sid) #can replace nb with np
```
So far I havent seen any cases where this fails, and it makes atleast the utilities in functions.py (while I had to copy and change the two lines to use the `charge` function with numba) numba friendly

eduardo-rodrigues:
Dear @TanmayPani, thank you for the interest in the package. Feel encouraged to say a bit more about your use case, with a code snippet where Particle is used, as it can in some cases make it into examples for future people ...

I agree that there is no string need to convert to a string. The function is largely the way it is for historical reasons. Do you fancy making a contribution with a PR? We always welcome that.

BTW, why the convertion with nb.int32 rather than simply returning sid?

TanmayPani:
Sorry for the slow reply @eduardo-rodrigues , Thanks for getting back! casting to `nb.int32` was mostly from my inexperience with numba, casting to `int` should work just  as well. I would be happy to contribute with adding  jitting support!

I use `particle` functions to calculate charge from pdg_id for particles event by event before inputting into  `fastjet.ClusterSequence` using the Awkward-Vector interface to get jets. I need the charge info to calculate stuff like jet-charge etc. from the jet-constituents later on. Here is snippets from my code:

```
import vector
import awkward as ak
from pdgid_utils import charge #copied and renamed 'particle/src/particle/pdgid/functions.py'
vector.register_awkward()
import numba as nb

@nb.jit
def charge_from_pdgid(builder, pid_arr):
    for pids in pid_arr:
[...]

PR #677, opened 2025-05-14
Improve helper PDGID related _digit function

eduardo-rodrigues:
Closes https://github.com/scikit-hep/particle/issues/671.

### DIFF
--- a/src/particle/pdgid/functions.py
+++ b/src/particle/pdgid/functions.py
@@ -901,7 +901,7 @@ def _digit(pdgid: PDGID_TYPE, loc: int) -> int:
 
     Returns the digit at position 'loc' >=0 given that the right-most digit is at position 1.
     """
-    assert loc > 0
+    assert loc > 0, "Internal inconsistency: `loc` is supposed to be >= 0, an attribute of `Location`"
     sid: float = (abspid(pdgid) % 10**loc) // (10 ** (loc - 1))
     return int(sid)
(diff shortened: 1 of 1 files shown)


src/particle/pdgid/functions.py
```diff
@@ -899,10 +899,11 @@ def _digit(pdgid: PDGID_TYPE, loc: int) -> int:
     """
     Provides a convenient index into the PDGID number, whose format is in base 10.
 
-    Returns the digit at position 'loc' given that the right-most digit is at position 1.
+    Returns the digit at position 'loc' >=0 given that the right-most digit is at position 1.
     """
-    sid = str(abspid(pdgid))
-    return int(sid[-loc]) if loc <= len(sid) else 0
+    assert loc > 0
```
jonas-eschle, line 904:
```suggestion
    assert loc > 0, "Internal inconsistency: `loc` is supposed to be >= 0, an attribute of `Location`"
```

jonas-eschle (approved):
Seems all good to me! I'd propose to add a message what the assert means and to make it clear that it's internal (i.e. no user error, a not-so-advanced user may doesn't understand the concept of assert vs exception).

src/particle/pdgid/functions.py
eduardo-rodrigues: That's a very good point, thanks.

PR #754, opened 2026-04-17
feat: drop Python 3.9 and start using pattern matching

henryiii:
This was testing a SKILL.md file (first one I've written) for dropping 3.9 (https://github.com/henryiii/skills/blob/main/drop_python3.9/SKILL.md). I was iterating on writing it, so if you rerun it it might behave a bit different (or using a different model, etc; I'm using nrp.ai, copilot and others support skills too).

:robot: Assisted-by: OpenCode: Kimi-K2.5

- **feat: drop Python 3.9**
- **chore: use pattern matching**

### DIFF
--- a/src/particle/pdgid/functions.py
+++ b/src/particle/pdgid/functions.py
@@ -657,39 +657,40 @@ def three_charge(pdgid: PDGID_TYPE) -> int | None:
     q3 = _digit(pdgid, Location.Nq3)
     sid = _fundamental_id(pdgid)
 
-    if _extra_bits(pdgid) > 0:
-        if is_nucleus(pdgid):  # ion
-            Z_pdgid = Z(pdgid)
-            return None if Z_pdgid is None else 3 * Z_pdgid
-        if is_Qball(pdgid):  # Qball
-            charge = 3 * ((aid // 10) % 10000)
-        else:  # this should never be reached in the present numbering scheme
-            return None  # since extra bits exist only for Q-balls and nuclei
-    elif is_dyon(pdgid):  # Dyon
-        charge = 3 * ((aid // 10) % 1000)
-        # this is half right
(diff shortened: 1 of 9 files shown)


src/particle/pdgid/functions.py
```diff
@@ -702,29 +703,27 @@ def j_spin(pdgid: PDGID_TYPE) -> int | None:
         return None
     if _fundamental_id(pdgid) > 0:
         fund = _fundamental_id(pdgid)
-        if is_SUSY(pdgid):  # susy particles
-            if 0 < fund < 17:
+        match (is_SUSY(pdgid), fund):
```
eduardo-rodrigues: The modernisation is nice (removal of Union, etc.) but the usage of case here does not clarify matters to me, to be honest. I find the code more obscure and complex. Ditto below. Do you find it easier to read?

eduardo-rodrigues: I mean, how is
```
match (is_SUSY(pdgid), fund):
    case (True, n) if 0 < n < 17:
```
easier than
```
        if is_SUSY(pdgid):  # susy particles
            if 0 < fund < 17:
```
for example?

And the comments got removed.

henryiii: Once you get used to reading match statements, they are nice for several reasons:
* Guaranteed to be about one variable, and it only gets evaluated once. If chains can change variables.
    * (Except for the guard clause, which is used too much here)
* Nice patterns for multiple values with `|`
* Can capture variables in the same place as it's checked

Two of these aren't better, though - if you use the if clause too much, then match statements aren't better. I'll leave the one that does and drop two of them.

eduardo-rodrigues: Sounds good, thank you.

eduardo-rodrigues (approved):

PR #756, opened 2026-04-24
Add support for RapidSim names

admorris:
I added support for the RapidSim particle-naming convention (oh how I wish they just used EvtGen names!)

It largely follows the same strategy as the LHCb names, but with the important difference of **loading from `$RAPIDSIM_ROOT/config/particles.dat` if present**, which is nice if you're using `particle` in e.g. a conda environment with `rapidsim` installed.


admorris:
To improve the coverage, I should add a test that mocks `$RAPIDSIM_ROOT` being set

eduardo-rodrigues:
Hello @admorris, I will look at this next week ...

One immediate reaction is indeed why RapidSim uses non PDG/EvtGen standard names in the 2020s!? It's more complication for absolute zero gain.

The obvious question is whether it would not be simpler in the end to sort the issue at source? In the end they must be using names as labels, right, so should be rather easy? Have you checked this with the authors?

I am not saying I am opposing to having another submodule as we already have for LHCb. Just asking for the record.

Thanks.

admorris:
Hi Eduardo

That's a fair suggestion.

I just had a go at converting RapidSim's `particles.dat` to use EvtGen names, and I found 4 instances where the PDG ID does not match :smiling_face_with_tear:  [particles.dat.txt](https://github.com/user-attachments/files/27057683/particles.dat.txt)

There are a lot of open PRs and issues in RapidSim which suggest it's not being actively maintained. But I'm sure Greig and Dan are still contactable.

eduardo-rodrigues:
Hi.

I checked briefly your data file and all 4 misses are for wrongly assigned PDGIDs from RapidSim authors. I could find the correct replacements in https://github.com/scikit-hep/particle/blob/main/src/particle/data/conversions.csv.

I don't know how you and LHCb would like to proceed, but that seems like the main point here, else I am happy to merge this as-is if that helps with support for LHCb.

I see many LHCb analysts still using RapidSim, hence it is useful and should be maintained. If the authors are not active anymore, could the package be taken over by LHCb somehow? I see it is effectively dead, since the last release is from 2019 and the latest merge from 2 years ago - https://github.com/gcowan/RapidSim!

admin/dump_pdgid_to_rapidsim.py
```diff
@@ -0,0 +1,65 @@
...
+    for pdgid_str, part_name, anti_name, *_ in table:
+        pdgid = int(pdgid_str)
+        rapidsim_names[pdgid] = part_name
+        if anti_name != "---":
+            rapidsim_names[-pdgid] = anti_name
+
+    with open("src/particle/rapidsim/data/pdgid_to_rapidsimname.csv", "w") as out_csv:
+        out_csv.write(
+            f"# (c) Scikit-HEP project - Particle package data file - pdgid_to_rapidsimname.csv - {date}\n"
+        )
```
eduardo-rodrigues, line 52:
```suggestion
        )
csvfile.write(
    "# Auto generated by 'admin/dump_pdgid_to_rapidsimname.py'\n",
        )
```

src/particle/rapidsim/data/pdgid_to_rapidsimname.csv
```diff
@@ -0,0 +1,185 @@
+# (c) Scikit-HEP project - Particle package data file - pdgid_to_rapidsimname.csv - 2026-04-24
```
eduardo-rodrigues, line 1:
```suggestion
# (c) Scikit-HEP project - Particle package data file - pdgid_to_rapidsimname.csv - 2026-04-24
# Auto generated by 'admin/dump_pdgid_to_rapidsimname.py
```

src/particle/rapidsim/converters.py
```diff
@@ -0,0 +1,85 @@
...
+    Parse a RapidSim ``particles.dat`` file and return a CSV-formatted
+    :class:`io.StringIO` with ``PDGID,STR`` rows suitable for :class:`BiMap`.
+
+    Each row in ``particles.dat`` provides a particle name and, when present
+    (i.e. not ``---``), an antiparticle name.  The antiparticle is mapped to
+    the negated PDG ID following the standard PDG sign convention.
+    """
+    lines: list[str] = ["PDGID,STR\n"]
+    with dat_file.open(encoding="utf-8") as f:
+        next(f)  # skip header line
```
eduardo-rodrigues, line 28: There is now an extra line to skip given my added comment, right?

eduardo-rodrigues (approved):
This looks great and I only made a couple of trivial suggestions.

Another suggestion would be to add somewhere in the docs the link to RapidSim, probably in admin/dump_pdgid_to_rapidsim.py but also in the functions to and from?

I let you decide if you update the docs for this new functionality here or in a follow-up PR.

Last thing: did you then decide what to do overall, I mean given the exchange above? Even if some work is going to be done to revive and modernise RapidSim, it may still take a while before that's available. If helpful we can move on merging and releasing this asap. Just let me know.

Thank you again for the great and clear enhancement.

eduardo-rodrigues:
Hi @admorris, I saw that RapidSim got moved to the LHCb organisation. Let me know your decision on this work.

admorris:
I had a quick chat with Michal K a few weeks ago about it, and I think we agreed it's better not to break backwards compatibility with the names, but also that it should be easier for the user to provide their own particle table. At the moment you need to copy the whole config directory and override the parts you need.

The PDG ID bugs should be fixed though IMO.

I am undecided on how that impacts whether this PR is needed. Someone who wants to use Particle is probably proactive enough to provide their own particle table to RapidSim (of course we could ship one with it and make it a flag).

eduardo-rodrigues:
OK, then we leave this open and see. Maybe you can chat to 1-2 usual suspects and see what their take is?

I agree that at least the PDG ID bugs ought to be fixed. Thanks.

eduardo-rodrigues:
Hi. I know this is in standby. Just to say, FYI, there's been in the past days a great enhancement that simplifies  conversions enormously, making them better designed, see https://github.com/scikit-hep/particle/pull/780.

eduardo-rodrigues:
Hello @admorris, kind ping on this PR.

My personal take is that it would be more beneficial to all if RapidSim (https://github.com/lhcb/RapidSim) would switch to using EvtGen names or PDG names, alternatively. The former makes even more sense since the docs tells me that there is now a connection to EvtGen and PHOTOS in the package; consistency is good.

Also, since the last release is from 7 years ago (sic!), a small breaking change would be no issue whatsoever. I'm not sure who the maintainer is. Can you ping them or point them to here?
Correction: let me just ping @michalkreps, @chrisburr and @leadreyfus since the history tells me they have been contributing much recently.

Thanks.

michalkreps:
> Hello @admorris, kind ping on this PR.
> 
> My personal take is that it would be more beneficial to all if RapidSim (https://github.com/lhcb/RapidSim) would switch to using EvtGen names or PDG names, alternatively. The former makes even more sense since the docs tells me that there is now a connection to EvtGen and PHOTOS in the package; consistency is good.
> 
> Also, since the last release is from 7 years ago (sic!), a small breaking change would be no issue whatsoever. I'm not sure who the maintainer is. Can you ping them or point them to here? Correction: let me just ping @michalkreps, @chrisburr and @leadreyfus since the history tells me they have been contributing much recently.
> 
> Thanks.

I agree that we should not add another convention here but rather update RapidSim to use convention already known here. I will be even bit stronger by saying that it is pity that EvtGen is using its own naming convention instead of using PDG names. Trouble with EvtGen is that it is breaking change so one has to be careful. I'm swamped right now, but will try to think about interplay of various packages in two or three weeks when I get bit more time.

eduardo-rodrigues:
Many thanks, @michalkreps. EvtGen names these days are everywhere and everyone accepted the differences wrt PDG names. Changing that indeed would be a massive thing with broad impact. Hence my suggestion ... For sure if you go for PDG names in RapidSim you will be screened from any breaking changes in EvtGen, so that would be the best course of action.

leadreyfus:
Hi @eduardo-rodrigues @michalkreps,

Just jumping in the conversation to add that Rapidsim's `particles.dat` is also somewhat incomplete (although it is probably enough for standard use): I wanted to generate some excited D states and I had to add the information by hand from what I could find in EvtGen's `evt.pdl`. Would it be possible to source the particle's info directly from evt.pdl maybe? The only additional information in `particles.dat` is the shape of the resonances, which is usually determined internally in EvtGen (see Sec. 12.2 https://evtgen.hepforge.org/doc/EvtGenGuide.pdf) unless specified otherwise in the decfile, if I understood correctly.

Overall, I do agree that changing the RapidSim convention to match the one of EvtGen is a good idea, I've had a few headaches with all of these excited D states names!

eduardo-rodrigues:
Hello @leadreyfus, good that you jumped in. Both routes are fine but IMO using PDG names is even a better move, as said.

As for your troubles, let me point out with a 100% bias that you could simply leverage this particles to get PDG, EvtGen names and all sorts of properties for free ;-). It's not a big dependency for a package to have. Just a thought.

PR #760, opened 2026-05-18
Implement baryon/lepton number and r-parity

### LINKED ISSUE #498: Tests for Lepton number, Baryon number and a function for R-parity
As per discussion in #494 with @eduardo-rodrigues 

- [ ] To write tests for lepton number and baryon number 
- [ ] Add a function to extract r-parity of susy particles

c.c.  @eduardo-rodrigues @redeboer

redeboer:
Closes #498


eduardo-rodrigues (commented):
Thank you very much for this @redeboer 👍!

I will try and review this asap now that I'm back from a conference ...

Before I do, I would like to know the opinion of @henryiii on the use of `fractions.Fraction`, which is not something we have been using so far, having 0.333... return elsewhere where 1/3 is returned.

@redeboer, as this is draft, is your intention to add some other bits? Or ...?

redeboer:
No worries, enjoy the conference!

> Before I do, I would like to know the opinion of @henryiii on the use of `fractions.Fraction`, which is not something we have been using so far, having 0.333... return elsewhere where 1/3 is returned.

Yep, fair point. I wonder if this is something that (if adopted) should be used elsewhere in the framework, too. In that case, it may lead to interface changes.

> @redeboer, as this is draft, is your intention to add some other bits? Or ...?

I removed the draft status.

eduardo-rodrigues:
Thanks. I am not at all familiar with Fraction. I wonder for a start if it complicates anything the user will have to do with retrieved "data" or algebraic calculations will be just as trivial as 3 * Fraction(...), meaning nothing is really imposed on the user.

redeboer:
Yeah in my experience it works just fine with algebraic operations. Could be that there are performance issues though?

henryiii:
It's much less performant, but someone can just cast it with `float()` to get back to floats. It probably doesn't support numba, etc, as well, but I think you can just call float. I'd be lightly biased to try it, I think.

eduardo-rodrigues:
Me too. But then we need to adapt the code elsewhere for places where we return charges in particular, which should then return a Fraction. I will try and find time ...

henryiii:
I can stick an AI on it, if you want.

eduardo-rodrigues:
Sure, it will certainly get at least most of the work done :D.

src/particle/particle/particle.py
```diff
@@ -772,6 +773,75 @@ def is_self_conjugate(self) -> bool:
...
+                return A
+            return -A
+        if self.pdgid.is_quark:
+            return Fraction(1, 3) if self.pdgid > 0 else Fraction(-1, 3)
+        return 0
+
+    @property
+    def lepton_number(self) -> int:
+        if self.pdgid.is_lepton:
+            return +1 if self.pdgid > 0 else -1
```
eduardo-rodrigues:
```suggestion
            return 1 if self.pdgid > 0 else -1
```

src/particle/particle/particle.py
```diff
@@ -772,6 +773,75 @@ def is_self_conjugate(self) -> bool:
...
+
+    @property
+    def r_parity(self) -> int | None:
+        if self.pdgid.J is None:
+            return None
+        B = self.baryon_number
+        L = self.lepton_number
+        s = self.pdgid.J
+        exponent = int(3 * B) + L + int(2 * s)
+        return +1 if exponent % 2 == 0 else -1
```
eduardo-rodrigues:
```suggestion
        return 1 if exponent % 2 == 0 else -1
```

eduardo-rodrigues:
@redeboer, I cannot finalise right now and will come back soon. One important thing to add is proper documentation, even if some of the functions are obvious. The documentation should also state the use of Fraction. Back soon.

redeboer:
Okay, done. But not sure how verbose those docstrings should be?
I wrote them in Sphinx style (which also works nicely in many editors), but I later that there is no documentation built with Sphinx :thinking:

src/particle/particle/particle.py
```diff
@@ -772,6 +773,150 @@ def is_self_conjugate(self) -> bool:
         """
         return self.anti_flag == Inv.Same
 
+    @property
+    def baryon_number(self) -> Fraction | int:
```
redeboer, line 778: #764 changes everything to `Fraction`. Perhaps that should happen here, too?

eduardo-rodrigues:
> Okay, done. But not sure how verbose those docstrings should be? I wrote them in Sphinx style (which also works nicely in many editors), but I later that there is no documentation built with Sphinx 🤔

Thanks a lot. I like "verbose" docstrings that explain the function/method and provide examples when things aren't utterly trivial. So I guess your verbosity is very welcome - will check.

Indeed we still do not have docs in Sphinx but that is something that would make sense at this point given that the package developed with several things. To be done ...

src/particle/particle/particle.py
eduardo-rodrigues: I'm confused. That's the case after this update of yours, right? So what else do you mean?

eduardo-rodrigues:
Hey @redeboer, am reviewing as I find a bit of time. I've just committed some little changes to the docstrings for consistency. Hope OK.

src/particle/particle/particle.py
redeboer: The order of merging those PRs doesn't matter to me. I just want the signatures/behavior to be consistent in the long run.

eduardo-rodrigues: Me too :). So we're good.

tests/particle/test_particle.py
```diff
@@ -783,3 +784,314 @@ def test_particle_hash(sign: int) -> None:
...
+        ("K+", Fraction(1, 2)),
+        ("K0", Fraction(-1, 2)),
+        # Baryons
+        ("p", Fraction(1, 2)),
+        ("n", Fraction(-1, 2)),
+        ("Lambda", 0),
+        ("Sigma+", 1),
+    ],
+)
+def test_gell_mann_nishijima(name: str, expected_i3_over_two: int | Fraction) -> None:
```
eduardo-rodrigues, line 1063:
```suggestion
def test_gell_mann_nishijima(name: str, expected_i3_over_two: int | Fraction) -> None:
    """Test the Gell-Mann–Nishijima formula relating charge to isospin and hypercharge."""
```

tests/particle/test_particle.py
```diff
@@ -783,3 +784,314 @@ def test_particle_hash(sign: int) -> None:
...
+    p = Particle.from_name(name)
+    Y = p.hypercharge
+    Q = p.charge
+    if Q is None:
+        pytest.skip(f"Particle {p.name} has no charge")
+    i3 = Q - Y / 2
+    assert i3 == expected_i3_over_two
+
+
+def test_gell_mann_nishijima_database_consistency() -> None:
```
eduardo-rodrigues, line 1074:
```suggestion
def test_gell_mann_nishijima_database_consistency() -> None:
    """Verify the Gell-Mann–Nishijima formula relating charge to isospin and hypercharge for all particles in the database."""
```

eduardo-rodrigues (approved):
Thank you very much again for this very nice enhancement, @redeboer 👍!

PR #763, opened 2026-06-01
chore: add support for agents

henryiii:
Adding an AGENTS.md file, which is supported by all harnesses _except_ Claude Code. Also adding CLAUDE.md to the ignore file so you can symlink CLAUDE.md to AGENTS.md to make Claude work too.

Assisted-by: ClaudeCode:claude-opus-4.8


eduardo-rodrigues (approved):
Fine by me.

PR #764, opened 2026-06-01
feat: return Fraction instead of float for charge, spin J, and isospin I

henryiii:
## Summary

Physics quantities that are inherently rational / half-integer now return an exact `fractions.Fraction` instead of a `float`, avoiding floating-point approximations (e.g. `Fraction(2, 3)` instead of `0.6666666666666666`).

Three quantities are converted:

- **Charge** — `pdgid.charge()` and the `PDGID.charge` / `Particle.charge` properties → `Fraction | None`. Quark/diquark charges are now exact (`2/3`, `-1/3`, …). `three_charge` stays `int` (already exact).
- **Total spin J** — `pdgid.J()` and the `PDGID.J` / `Particle.J` properties → `Fraction | None` (e.g. electron `1/2`, Δ baryon `3/2`). The incorrect `-> int` annotation on `Particle.J` is corrected, and `Particle.spin_type` now derives the integer J directly from `j_spin`.
- **Isospin I** — `Particle.I` and `_isospin_converter` → `Fraction | None`. The source data already stores these as `"1/2"`, `"3/2"`, … strings, so the mapping is natural.

`S`, `L` (meson spin / orbital angular momentum) and the `j_spin`/`s_spin`/`l_spin` helpers remain `int` — they are genuine integers and gain nothing from `Fraction`.

## Behavior changes

- The above properties/functions return `Fraction` rather than `float`. `Fraction` interoperates with numeric comparisons (`charge == 1`, `I == 0.5`, `J != 0` still work), but exact equality against a non-dyadic `float` literal like `-2/3` no longer holds — compare against `Fraction(-2, 3)` instead.
- Display changes (exact values, no trailing `.0`):
  - `PDGID.info()` shows `charge  0` and `J  1` rather than `0.0` / `1.0`.
  - `Particle.describe()` / `_str_charge()` show nucleus charges and `J`/`I` as exact values (e.g. `1`, `1/2`, `3/2`).

## Tests

- Updated `tests/pdgid/test_functions.py` (fractional charge vs `Fraction(-2, 3)`), `tests/pdgid/test_pdgid.py` (`info()` display), and `tests/particle/test_particle.py` (isospin checklist uses `Fraction`).
- Full suite: 399 passed; `prek -a` (ruff, mypy strict, …) clean.


henryiii:
:robot: Which of these should I also convert to Fraction in this PR?

1. [x] Total spin J
  pdgid.J / Particle.J return Fraction (0.5, 1.5, ...) instead of float. Also fixes the incorrect `-> int` annotation on Particle.J.
2. [x] Isospin I
  Particle.I returns Fraction (1/2, 3/2, ...) instead of float; _isospin_converter maps to Fraction.

eduardo-rodrigues:
Yeah, then we should be consistent, even if 0.5 and alike to not in general need Fraction since given by a finite number of digits. I guess you @redeboer also did the same in your package, right?

redeboer:
> I guess you @redeboer also did the same in your package, right?

Yap https://github.com/ComPWA/qrules/pull/288

eduardo-rodrigues:
@henryiii, I will try and look at this very soon. Just one think that came to my mind on the spot - did you check if any of the docs needs to be updated for these API changes?

eduardo-rodrigues (approved):
Replying to myself - this PR is reviewed and seen ready aside minor doc updates in the README.

tests/particle/test_particle.py
```diff
@@ -590,29 +591,29 @@ def test_spin_type(pid: int, stype: SpinType) -> None:
...
+    (-421, Fraction(1, 2)),  # D0_bar
+    (333, Fraction(0)),  # phi(1020)
+    (443, Fraction(0)),  # J/psi
+    (521, Fraction(1, 2)),  # B+
+    (531, Fraction(0)),  # Bs
     # Baryons
-    (2212, 0.5),  # proton
-    (2214, 1.5),  # Delta+
+    (2212, Fraction(1, 2)),  # proton
+    (2214, Fraction(3, 2)),  # Delta+
```
redeboer: For non-fractional charges, one can just compare with an integer, e.g.:

```suggestion
    (211, 1),  # pi+
    (310, Fraction(1, 2)),  # K_S
    (-421, Fraction(1, 2)),  # D0_bar
    (333, 0),  # phi(1020)
    (443, 0),  # J/psi
    (521, Fraction(1, 2)),  # B+
    (531, 0),  # Bs
# Baryons
    (2212, Fraction(1, 2)),  # proton
    (2214, Fraction(3, 2)),  # Delta+
```

In fact, comparing `Fraction(1, 2)` with `0.5` also works.

redeboer (approved):
Looks good, just a style thing in the tests I was wondering about.

tests/particle/test_particle.py
eduardo-rodrigues: That's even nice to have as you suggest as it immediately shows that the comparison with floats work. I would accept this.

eduardo-rodrigues:
> Replying to myself - this PR is reviewed and seen ready aside minor doc updates in the README.

I went ahead myself, @henryiii, see https://github.com/scikit-hep/particle/pull/764/commits/82ee26f43657a8f92210fe9c7e2f2e604d6c7548. Approving.

eduardo-rodrigues (approved):

tests/particle/test_particle.py
```diff
@@ -590,29 +591,29 @@ def test_spin_type(pid: int, stype: SpinType) -> None:
...
     (-12, None),  # nu(e)_bar
     # Mesons
-    (211, 1.0),  # pi+
-    (310, 0.5),  # K_S
-    (-421, 0.5),  # D0_bar
-    (333, 0.0),  # phi(1020)
-    (443, 0.0),  # J/psi
-    (521, 0.5),  # B+
-    (531, 0.0),  # Bs
+    (211,1),  # pi+
```
eduardo-rodrigues, line 602:
```suggestion
    (211, 1),  # pi+
```

tests/particle/test_particle.py
eduardo-rodrigues: Done manually as I was getting an error applying the suggestion.

eduardo-rodrigues:
Let me merge this so that `fractions.Fraction` gets introduced before the enhancement https://github.com/scikit-hep/particle/pull/760 goes in, which is where `fractions.Fraction` was originally proposed.

Issue #772, opened 2026-06-11
Code review: bugs, performance, simplifications, and modernizations


henryiii:
:robot: _AI text below_ :robot:

A full review of the codebase (every source module, packaging config; suspicious findings verified by execution) turned up the items below, grouped by category. Draft PRs addressing them will reference this issue.

## Bugs

- [ ] **`Particle.to_list` / `to_dict` silently ignore `search_terms` and `particle=` when `filter_fn` is not given** — `src/particle/particle/particle.py:414-416`. The search is only applied `if filter_fn is not None`. Confirmed: `Particle.to_list(pdg_name='tau')` and `Particle.to_list(particle=True)` both return the full table (6506 rows). Fix: call `cls.findall(filter_fn, particle=particle, **search_terms)` unconditionally; this also removes the redundant `sorted(cls.all())` (the table is already sorted by `load_table`, and that sort is computed then discarded whenever a filter *is* passed).
- [ ] **CLI rough edges** (`src/particle/__main__.py`): `python -m particle` with no subcommand prints nothing and exits 0 (subparsers should be required, as `convert.py` does); `python -m particle pdgid abc` and `python -m particle search 99999999` die with raw tracebacks instead of a friendly message + exit 1.
- [ ] **`from_name("p")` can raise bare `StopIteration`** — `particle.py:1185`. The p/n/p~/n~ branch uses `next(filter(...))`; with a user-replaced table lacking those entries this leaks `StopIteration` instead of `ParticleNotFound`.
- [ ] **`str_with_unc` crashes on `value == 0` with nonzero uncertainty** — `src/particle/particle/utilities.py:71`, `math.log10(value)` → `ValueError: math domain error` (verified). No shipped table entry triggers it, but a user-loaded table with a 0 ± ε mass would crash `describe()`/`repr()`.
- [ ] **Error-message/doc typos**: `particle.py:1259` Z validation says "Must be 0 <= A <= 999"; `particle.py:1271` "Z must be smaller or equal to Z"; `kinematics.py:53,99` raise on `< 0` but say `<= 0`; `lifetime_to_width` "Returns" docstring describes a lifetime in ns instead of a width in MeV; `converters/evtgen.py` `EvtGen2PDGNameMap` docstring writes the expected output as a `>>>` prompt line; `geant3id.py:52` comment says "Keep PythiaID..." (copy-paste).
[...]

PR #773, opened 2026-06-11
fix: to_list/to_dict ignored search terms when no filter_fn given

henryiii:

## Summary

**Bug:** `Particle.to_list(**search_terms)` and `Particle.to_list(particle=...)` silently returned the entire table (6506 rows) when `filter_fn` was not given, because `search_terms` and `particle` were only forwarded to `findall` inside an `if filter_fn is not None:` guard.

**Before:**
```python
>>> Particle.to_list(pdg_name="tau", exclusive_fields=["pdgid"])
# returned all 6506 particles, not just tau/anti-tau
```

**After:**
```python
>>> Particle.to_list(pdg_name="tau", exclusive_fields=["pdgid"])
[['pdgid'], [15], [-15]]  # only the two tau rows
```

**Fix:** Call `cls.findall(filter_fn, particle=particle, **search_terms)` unconditionally. With all-None/empty arguments `findall` returns the full table, preserving the no-filter behavior. The redundant `sorted(cls.all())` (computed then discarded whenever a filter was present) is eliminated as a side effect.

Two regression tests are added:
- `test_to_list_search_terms_without_filter_fn`: asserts `pdg_name="tau"` yields exactly the header + tau (15) + anti-tau (−15).
- `test_to_list_particle_kwarg_without_filter_fn`: asserts `particle=True` yields only positive PDG IDs.

Addresses one item of #772

### DIFF
--- a/tests/particle/test_particle.py
+++ b/tests/particle/test_particle.py
@@ -664,7 +664,7 @@ def test_to_dict() -> None:
 
 
 def test_to_list_search_terms_without_filter_fn() -> None:
-    # Regression: search_terms were silently ignored when filter_fn was not given.
+    # Test the search when filter_fn is not given.
     tbl = Particle.to_list(pdg_name="tau", exclusive_fields=["pdgid"])
     # Header row + tau (15) + anti-tau (-15)
     assert len(tbl) == 3
@@ -674,7 +674,7 @@ def test_to_list_search_terms_without_filter_fn() -> None:
 
 
 def test_to_list_particle_kwarg_without_filter_fn() -> None:
-    # Regression: the particle= keyword was silently ignored when filter_fn was not given.
+    # Test that the particle= keyword is not silently ignored when filter_fn is not given.
     tbl = Particle.to_list(particle=True, n_rows=5, exclusive_fields=["pdgid"])
     pdgids = [int(row[0]) for row in tbl[1:]]  # skip header
     assert all(pid > 0 for pid in pdgids)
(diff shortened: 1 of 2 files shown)


tests/particle/test_particle.py
```diff
@@ -663,6 +663,23 @@ def test_to_dict() -> None:
     assert set(query_as_dict["name"]) == {"e+", "mu+", "tau+", "tau'+"}
 
 
+def test_to_list_search_terms_without_filter_fn() -> None:
+    # Regression: search_terms were silently ignored when filter_fn was not given.
```
eduardo-rodrigues:
```suggestion
    # Test the search when filter_fn is not given.
```

tests/particle/test_particle.py
```diff
@@ -663,6 +663,23 @@ def test_to_dict() -> None:
...
+    tbl = Particle.to_list(pdg_name="tau", exclusive_fields=["pdgid"])
+    # Header row + tau (15) + anti-tau (-15)
+    assert len(tbl) == 3
+    assert ["pdgid"] in tbl
+    assert [15] in tbl
+    assert [-15] in tbl
+
+
+def test_to_list_particle_kwarg_without_filter_fn() -> None:
+    # Regression: the particle= keyword was silently ignored when filter_fn was not given.
```
eduardo-rodrigues:
```suggestion
    # Test that the particle= keyword is not silently ignored when filter_fn is not given.
```

eduardo-rodrigues (approved):
Merging, good catch!

PR #776, opened 2026-06-11
perf: single-pass BiMap loading and O(1) generator-ID reverse lookups

henryiii:

## Summary

Addresses converter performance items of #772.

- **`BiMap` / `DirectionalMaps` single-pass loading** (`src/particle/converters/bimap.py`): both constructors previously made two passes over the CSV — building `_to_map`, seeking to position 0, then building `_from_map`. Refactored to a single `for` loop that fills both dicts simultaneously. Side-effect: removes the latent crash when a caller passes a non-seekable file object (e.g. a `BytesIO` or stdin-derived stream).

- **O(1) `from_pdgid` lookups** (`pythiaid.py`, `geant3id.py`, `corsika7id.py`): each `from_pdgid` classmethod previously did a linear scan over `_bimap` (up to ~250 entries for Pythia). Added a module-level `_inverse_bimap` built at import time with `setdefault` to preserve first-match-wins semantics consistent with the original `for` loop, then replaced the scan with a single dict lookup.

## Duplicate PDGID check

All three CSVs were checked for duplicate PDGID values before building the inverse map:

| File | Duplicate PDGIDs |
|------|-----------------|
| `pdgid_to_pythiaid.csv` | None |
| `pdgid_to_geant3id.csv` | None |
| `pdgid_to_corsika7id.csv` | None |

No duplicates found in any file, so the `setdefault` guard is a safety measure only — it has no effect on the current data.

## Timing

Measured `PythiaID.from_pdgid(9010221)` (a late entry in the map, worst-case for the old linear scan) with `timeit` over 10 000 iterations:

| | µs/call |
|---|---|
| Before | ~5.88 µs |
| After | ~0.06 µs |

**~100× speedup** for a single call; scales linearly with call volume.

## Test plan

- [x] `uv run pytest` — 550 passed, 2 skipped (pandas not installed, unrelated)
- [x] `prek -a --quiet` — all hooks pass (ruff, mypy strict, codespell, etc.)
- [x] All converter, corsika, geant, and pythia test modules pass
- [x] Behavior-preserving: no API changes


eduardo-rodrigues:
Hello @The-Ludwig, this impressive performance enhancement touches a lot on code that you also touch on your redesign at https://github.com/scikit-hep/particle/pull/780. This is bound to provide some conflicts for you. Is it still OK to you that I go ahead, review and most likely merge? And you then adapt your PR https://github.com/scikit-hep/particle/pull/780? I will wait for your opinion on this to avoid unforeseen pain.

The-Ludwig:
> This is bound to provide some conflicts for you. Is it still OK to you that I go ahead, review and most likely merge?

Yes, absolutely. Claude came up with essentially the same solution in my PR, so it's already in there; Only difference is that my PR builds it lazily upon first call and doesn't build the reverse map on the `csv` read. But I'm happy to adapt #780!

eduardo-rodrigues:
OK, super, many thanks for the flexibility/willingness 👍.

eduardo-rodrigues (approved):
Great improvement!

eduardo-rodrigues:
OK, @The-Ludwig, I merged this now. I  think you can go ahead with your enhancement :). But do enjoy the weekend first ...

The-Ludwig:
Hi @eduardo-rodrigues ,
I enjoyed my weekend, I hope you did as well.
I incorporated the changes from this PR into #780 , but as said above this deletes all changed for the different particle ID classes,  as the O(1) lookup was part of the new MCParticleID base-class rewrite.

One more thing: I personally would say that the 1.0 release is a good point to just deprecate the `BiMaps` for the particle ID classes. Ultimately that choice is up to you, the maintainers! :)

eduardo-rodrigues:
Yeah, nice weekend, thanks.

Thank you also for the comment. I tend to agree with you; just let me check the changes in your PR, though not anymore today.

PR #777, opened 2026-06-11
perf: hoist per-call table constants in pdgid/functions.py

henryiii:

## Summary

Four behaviour-preserving performance/simplification refactors in `src/particle/pdgid/functions.py`. Addresses several performance/simplification items of #772.

### Changes

1. **`three_charge` — hoist `ch100` to module-level `_CH100` tuple**
   The 100-element list was rebuilt from literals on *every* call. This is on the hot path for all charge lookups, including every `Particle.findall` filter that touches charge.
   - Before: ~2.17 µs/call
   - After: ~2.05 µs/call (~6% faster, ~120 ns/call saved)

2. **`has_fundamental_anti` — hoist `_cp_conjugates` and `_unassigned` to module-level frozensets**
   Both were rebuilt per call; `_unassigned` was a list, so membership tests were O(n) linear scans. Now both are module-level `frozenset` constants for O(1) lookup. As a bonus, the duplicate `26` entry (present both as a literal and via `range(26, 32)`) is naturally deduped — semantics unchanged: `{9, 10, 19, 20} | set(range(26, 32)) | set(range(45, 80))`.

3. **`is_pentaquark` — merge two separate `Nj` checks into one**
   `_digit(pdgid, Location.Nj) == 9` was checked at one guard and `== 0` at a later guard. Merged into a single `if _digit(pdgid, Location.Nj) in {0, 9}: return False`, saving one redundant call to `_digit`.

4. **`j_spin` — collapse consecutive SUSY branches**
   `if fund == 21: return 2` followed by `if 22 <= fund < 38: return 2` both return 2, collapsed to `if 21 <= fund < 38: return 2`.

## Test plan

- [x] `uv run pytest tests/pdgid/` — 48 tests, all passed
- [x] `uv run pytest` — 561 tests, all passed
- [x] `prek -a --quiet` — ruff, mypy (strict), codespell all clean

### DIFF
--- a/src/particle/pdgid/functions.py
+++ b/src/particle/pdgid/functions.py
@@ -49,6 +49,119 @@ class Location(IntEnum):
     N10 = 10
 
 
+# Maps fundamental PDG ID (1..100) to 3*charge.  Index i corresponds to fundamental ID i+1.
+_CH100: tuple[int, ...] = (
+    -1,
+    2,
+    -1,
+    2,
+    -1,
+    2,
+    -1,
+    2,
+    0,
+    0,
+    -3,
+    0,
+    -3,
+    0,
+    -3,
+    0,
+    -3,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    3,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    3,
+    0,
+    0,
+    3,
+    0,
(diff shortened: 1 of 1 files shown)


eduardo-rodrigues (approved):
This brings in small but nice simplifications, and the non-negligible speedup is also welcome.

@henryiii, merging as I assume you are happy also with this simple update.

PR #778, opened 2026-06-11
perf: make particle and pdgid literals lazy to avoid loading the table at import

henryiii:

## Problem

`import particle` was eagerly loading the full particle CSV table, contradicting the documented lazy-loading design (the comment in `src/particle/__init__.py` and `CLAUDE.md` both state the CSV is not read until a particle is accessed).

`python -X importtime` traced ~27 ms to `particle.particle.literals` → `particle.shared_literals`. The chain:

- `particle/particle/literals.py` and `particle/pdgid/literals.py` (cog-generated) imported `shared_literals` and ran ~1000 eager module-level `Particle.from_pdgid(...)` / `PDGID(...)` statements at import.
- `shared_literals.common_particles` is built via `Particle.findall(...)`, which forces the full table load.

## Fix (PEP 562 lazy module attributes)

- Cog now emits a static `common_particles: dict[str, int]` (name → PDG ID), frozen at generation time, into **both** literals modules — neither imports `shared_literals` at runtime anymore.
- Each module defines a module-level `__getattr__(name)` that looks the name up in the static dict, constructs the `Particle` / `PDGID` on first access, caches it in `globals()`, and raises a helpful `AttributeError` for unknown names. `__dir__` still lists all names for tab completion.
- `shared_literals.py` remains the generation-time source of truth used by cog / `nox -s generate_aliases`, but is off the runtime import path.
- `uv run nox -s generate_aliases` regenerates exactly what is checked in (cog reports no change on re-run; lint is stable).

Note: PDGID literals construction is itself cheap (no table needed); accessing them no longer triggers a table load either, since `shared_literals` is gone from the import path.

## Proof: table not loaded at import

```
$ uv run python -c "import particle; print(particle.Particle.table_loaded())"
False
$ uv run python -c "from particle import literals as lp; print(lp.pi_plus); import particle; print(particle.Particle.table_loaded())"
pi+
True
$ uv run python -c "from particle.literals import Lambda_b_0; print(Lambda_b_0)"
Lambda(b)0
$ uv run python -c "from particle.pdgid import literals; print(literals.pi_plus)"
<PDGID: 211>
```

## Import time (before / after)

`uv run python -X importtime -c "import particle" 2>&1 | tail -1` (cumulative, last line):

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| **before** (main) | 59.9 ms | 56.0 ms | 55.5 ms |
| **after** | 27.1 ms | 25.5 ms | 26.0 ms |

Roughly a 2x faster import.

## Tests / lint

- `uv run pytest`: 561 passed.
- `prek -a --quiet`: clean (ruff, mypy strict, codespell, etc.). Module-level `__getattr__` is typed `-> Particle` / `-> PDGID`.
- `uv run nox -s generate_aliases`: no diff.

Addresses the main performance item of #772.


eduardo-rodrigues (approved):
This is a very good performance improvement. All looks good to me. I'm merging, @henryiii, but do have a look just in case; thank you.

PR #779, opened 2026-06-11
fix: assorted small correctness fixes (from_name, str_with_unc, load_table, messages)

henryiii:

## Summary

Addresses several small bug items of #772.

- **`from_name` StopIteration → ParticleNotFound**: the special branch for `{"p","n","p~","n~"}` used bare `next(filter(...))` which raises `StopIteration` instead of `ParticleNotFound` when those entries are absent from a user-replaced table. Fixed to use `next(..., None)` with an explicit raise.
- **`str_with_unc` zero-value crash**: `math.log10(0)` raised `ValueError: math domain error` when `value == 0` with a nonzero uncertainty (reachable via `describe()`/`repr()` on user-table particles). Added an early guard that formats using only the error's significant digits.
- **`load_table(append=True)` double-read**: calling `load_table(append=True)` with `filename=None` on a fresh unloaded table read the default CSVs twice (four `table_names` entries instead of two). Fixed by returning early after the recursive default load when `filename is None`.
- **`from_nucleus_info` error messages**: Z-range check said `"Must be 0 <= A <= 999"` (should be Z); the `z > a` check said `"Z must be smaller or equal to Z"` (should be A).
- **`kinematics.py` error messages and docstring**: `ValueError` raised on `< 0` input but messages said `<= 0` (0 is valid, returning `inf`). Also fixed the `lifetime_to_width` Returns docstring that was copy-pasted from `width_to_lifetime` and described a lifetime instead of a width.
- **`evtgen.py` docstring**: `EvtGen2PDGNameMap.__doc__` example had `>>> 'J/psi(1S)'` — the expected output line incorrectly had `>>>`. Removed it.
- **`geant3id.py` comment**: `# Keep PythiaID a slots based class` was a copy-paste error; corrected to `Geant3ID`.

Tests added for `str_with_unc(0.0, 1e-9, 1e-9)`, `from_name` raising `ParticleNotFound` on the special names when the table lacks those entries, and `load_table(append=True)` not duplicating `table_names`.

### DIFF
--- a/src/particle/converters/evtgen.py
+++ b/src/particle/converters/evtgen.py
@@ -51,7 +51,7 @@ Directional map between EvtGen and names.
 Examples
 --------
 >>> EvtGen2PDGNameMap['J/psi']
->>> 'J/psi(1S)'
+'J/psi(1S)'
 """
 
 del BiMap
--- a/src/particle/geant/geant3id.py
+++ b/src/particle/geant/geant3id.py
@@ -47,7 +47,7 @@ class Geant3ID(int):
     'pi+'
     """
 
-    __slots__ = ()  # Keep PythiaID a slots based class
+    __slots__ = ()  # Keep Geant3ID a slots based class
 
     @classmethod
     def from_pdgid(cls: type[Self], pdgid: int) -> Self:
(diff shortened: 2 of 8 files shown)


eduardo-rodrigues (approved):
Various improvements and documentation fixes. Extra tests are also good to have.

PR #780, opened 2026-06-11
feat: common MCParticleID base class for MC particle IDs

### LINKED ISSUE #427: Idea: Generalize Particle ID Classes and Improve
This idea came up when I worked on [implementing Corsika7 IDs](https://github.com/scikit-hep/particle/pull/426). 

Right now, every MC Particle ID class (`PDGID`, `Geant3ID`, `PythiaID`) is independent from each other, while being very similar in the sense that they inherit from `int`. 
I think it might be useful to define some kind of common behavior of  the particle IDs, to streamline the use of other particle IDs, something like:
```python
from abc import ABC, abstractmethod

class ForeignParticleID(ABC, int):
    @abstractmethod
    def to_pdgid(self) -> PDGID:
        pass
    
    @abstractmethod
    def from_pdgid(cls: Self, pdgid: PDGID) -> Self:
        pass
```
Of course, this conversion might not always be possible, in which case a `NoMatchingID` error is raised. 

This could then also replace the converter `BiMap` objects, I think they are not the optimal. Especially the `Pythia2PDGIDBiMap` is funny, since PythiaID and PDGID are the same, if existing. There should be converters which can be customly defined, not relying on a csv file.

The-Ludwig:
Edit:

- New commits with old, boring Opus 4.8... God what crazy times. I have to say Fable-5 was so much better. 
- Added back the `BiMap` classes and added a deprecation wrapper upon first use to them
- Incorporated changes from #776 . Most of that PR got deleted for the ParticleID classes because this already implemented a O(1)-lookup. 
- Note that this does not preserve the first-match-wins behavior for the conversion, but has last-match-wins behavior. Both should never happen and are arguably equally valid.

Hi! I was searching for a good test-case for anthropic's Fable-5, while I'm still able to use it. This came up again at the right time, and since this repository seems to accept LLM-assisted code, I thought I already give it a try. 
From my judgement, the changes below are all high-quality and very deliberate in solving the idea outlined in #427 . Some things were implemented I honestly wouldn't have noticed myself (for instance I learned about `slot-based classes`). 
Following is a claude-generated, more detailed description of the changes.

Closes #427.

## Summary

The MC particle ID classes (`Geant3ID`, `PythiaID`, `Corsika7ID`) were independent `int` subclasses with near-identical, duplicated machinery, while the `BiMap` converter objects re-read the same conversion CSVs a second time at import. This PR introduces a common base class, `particle.mcid.MCParticleID`, that defines the shared contract:

- subclasses define a class-level `_to_pdg_map` mapping (ID → PDG ID) and inherit `to_pdgid()`/`from_pdgid()` for free, **or** override both methods when the conversion is algorithmic and no table is needed;
- failed conversions uniformly raise `MatchingIDNotFound`;
- the inverse (PDG ID → ID) map is built lazily once per class, making `from_pdgid` O(1) instead of the previous linear scan over the table;
- downstream users can define custom ID classes for their own generators without any CSV file — a class body with an inline dict suffices.

`MCParticleID` is deliberately *not* an ABC: the abstract-instantiation check lives in `object.__new__`, which `int.__new__` bypasses, so `@abstractmethod` on an `int` subclass is silently unenforced. Instead, a subclass that neither defines a map nor overrides the methods gets a `NotImplementedError` with instructions at first use. It is also deliberately not exported at the top level (subclassing it is a rare, advanced use case) — import it from `particle.mcid`, analogous to `BiMap` in `particle.converters.bimap`.

## PythiaID: conversion table and whitelist removed

Analysis of `pdgid_to_pythiaid.csv` showed that of its 541 rows, **536 were identity mappings**. The entire information content was five light scalar mesons for which Pythia kept older PDG numberings:

| PDG ID | Pythia ID | particle |
|---|---|---|
| 10221 | 10331 | f₀(1370) |
| 9000111 | 10111 | a₀(980)⁰ |
| ±9000211 | ±10211 | a₀(980)± |
| 9010221 | 10221 | f₀(980) |

`PythiaID` conversions are therefore now algorithmic — identity plus this hardcoded 5-entry exception dict — and the CSV file is deleted.

Don't take my word for the completeness of the five exceptions — this snippet re-derives the dict from the old CSV in git history and compares:

```python
import csv, io, subprocess

from particle.pythia.pythiaid import _pdgid_to_pythiaid

old_csv = subprocess.run(
    ["git", "show", "master:src/particle/data/pdgid_to_pythiaid.csv"],
    capture_output=True, text=True, check=True,
).stdout
rows = csv.DictReader(line for line in io.StringIO(old_csv) if not line.startswith("#"))
non_identity = {
    int(row["PDGID"]): int(row["PYTHIAID"])
    for row in rows
    if int(row["PDGID"]) != int(row["PYTHIAID"])
}

print(f"non-identity rows in the old CSV: {non_identity}")
assert non_identity == _pdgid_to_pythiaid
```

**Deliberate behavior change:** the CSV also implicitly acted as a 541-particle whitelist of "IDs Pythia knows about", so `PythiaID.from_pdgid()` used to raise `MatchingIDNotFound` for any PDG ID outside it. Now such IDs convert via identity. The whitelist was a stale Pythia-6-era snapshot: Pythia 8 follows the PDG numbering scheme and knows far more than 541 particles, so the package was pretending to validate Pythia's particle content with data that was wrong (merely wrong in the conservative direction, rejecting valid conversions). Maintaining that list had no upside; dropping it makes the class honest about what it can guarantee.

## Other behavior changes

- **Removed without deprecation:** `Pythia2PDGIDBiMap`, `Geant2PDGIDBiMap`, `Corsika72PDGIDBiMap` and their modules. They duplicated what the ID classes themselves now do via `to_pdgid()`/`from_pdgid()`. The `BiMap`/`DirectionalMap(s)` classes and the string-based EvtGen/LHCb name maps are untouched. Happy to convert this into a two-step deprecation (keep the three BiMaps emitting `DeprecationWarning`, built from the class maps) if preferred, as offered in #427.
- **`Geant3ID.to_pdgid()`** now raises `MatchingIDNotFound` for an unknown ID instead of leaking a bare `KeyError`.
- **`Corsika7ID.to_pdgid()`** semantics sharpened: `InvalidParticle` is raised only when the ID does not correspond to a particle at all (bookkeeping entries, invalid IDs), while a real particle with no PDG match — an unmapped nucleus in the 200–5699 range — now raises `MatchingIDNotFound` (previously also `InvalidParticle`). This matches the validity logic of `from_particle_description`.

## Also in this PR

- Module-level imports of `Particle`/`InvalidParticle` in `corsika7id.py`: the in-function imports worked around a circular import through `converters/corsika.py`, which no longer exists.
- New `tests/test_mcid.py` covering the base class directly; ID-class tests extended for the new exception paths and all five Pythia legacy numberings; README converter section rewritten around the ID classes, including a custom-subclass example.

CHANGELOG deliberately untouched, following the convention of updating it in release-preparation PRs.

🤖 Co-generated with [Claude Code](https://claude.com/claude-code)


eduardo-rodrigues:
Hello @The-Ludwig, a lot has been happening this week, for good reasons, but it did prevent me from having much time to review this enhancement from you. I just wanted to say that I thank you already for this and it is not forgotten, back soon here :).

I would also welcome @henryiii's eyes in this, hence the review request.

src/particle/converters/bimap.py
```diff
@@ -135,6 +136,43 @@ def __len__(self) -> int:
...
 
+class _DeprecatedBiMap(BiMap[A, B]):
+    """
+    A :class:`BiMap` that emits a ``DeprecationWarning`` on first use.
+
+    Used for the PDG-ID <-> MC-program-ID maps, which are deprecated in
+    favour of the ``from_pdgid()`` and ``to_pdgid()`` methods of the ID
+    classes themselves. The warning is emitted only once (on the first
+    look-up) to avoid spamming; the map behaves identically to
+    :class:`BiMap` otherwise.
```
eduardo-rodrigues, line 147:
```suggestion
    :class:`BiMap` otherwise.
    
    This helper function will be removed at a later stage, in the medium term.
```

src/particle/converters/corsika.py
```diff
@@ -8,19 +8,22 @@
...
 from ..pdgid import PDGID
-from .bimap import BiMap
+from .bimap import _DeprecatedBiMap
 
-Corsika72PDGIDBiMap = BiMap(PDGID, Corsika7ID)
+Corsika72PDGIDBiMap = _DeprecatedBiMap(PDGID, Corsika7ID, name="Corsika72PDGIDBiMap")
 Corsika72PDGIDBiMap.__doc__ = """
 Bi-bidirectional map between PDG and Corsika7 IDs.
 
+.. deprecated::
```
eduardo-rodrigues:
```suggestion
.. deprecated:: 1.0.0
```

src/particle/converters/geant.py
```diff
@@ -8,19 +8,22 @@
...
 from ..pdgid import PDGID
-from .bimap import BiMap
+from .bimap import _DeprecatedBiMap
 
-Geant2PDGIDBiMap = BiMap(PDGID, Geant3ID)
+Geant2PDGIDBiMap = _DeprecatedBiMap(PDGID, Geant3ID, name="Geant2PDGIDBiMap")
 Geant2PDGIDBiMap.__doc__ = """
 Bi-bidirectional map between PDG and Geant3 IDs.
 
+.. deprecated::
```
eduardo-rodrigues:
```suggestion
.. deprecated:: 1.0.0
```

src/particle/converters/pythia.py
```diff
@@ -8,19 +8,22 @@
...
 from ..pythia import PythiaID
-from .bimap import BiMap
+from .bimap import _DeprecatedBiMap
 
-Pythia2PDGIDBiMap = BiMap(PDGID, PythiaID)
+Pythia2PDGIDBiMap = _DeprecatedBiMap(PDGID, PythiaID, name="Pythia2PDGIDBiMap")
 Pythia2PDGIDBiMap.__doc__ = """
 Bi-bidirectional map between PDG and Pythia IDs.
 
+.. deprecated::
```
eduardo-rodrigues:
```suggestion
.. deprecated:: 1.0.0
```

src/particle/corsika/corsika7id.py
```diff
@@ -209,16 +195,15 @@ def to_pdgid(self) -> PDGID:
...
         InvalidParticle: The Corsika7ID <Corsika7ID: 76> does not correspond to a particle and thus has no equivalent PDGID.
         """
-        from ..particle.particle import InvalidParticle  # pylint: disable=C0415
+        if int(self) in self._to_pdg_map:
+            return super().to_pdgid()
 
-        if self not in _bimap:
-            raise InvalidParticle(
-                f"The Corsika7ID {self} does not correspond to a particle and thus has no equivalent PDGID."
+        # Nuclei with no known PDG ID are particles nonetheless
```
eduardo-rodrigues: Is this something overlooked in the past?
Take one example, corresponding in Corsika7 to 4818:
```
In [3]: Particle.from_pdgid(1000180480)
Out[3]: <Particle: name="Ar48", pdgid=1000180480, mass=44689.3620515 MeV>
```

There is similar code above in the untouched part.

src/particle/mcid.py
```diff
@@ -0,0 +1,124 @@
...
+from .exceptions import MatchingIDNotFound
+from .pdgid import PDGID
+
+Self = TypeVar("Self", bound="MCParticleID")
+
+
+def _csv_to_pdg_map(filename: str, id_column: str) -> dict[int, int]:
+    """
+    Read an <ID> -> PDG ID conversion table from a CSV file
+    shipped in the package data.
```
eduardo-rodrigues, line 26:
```suggestion
    shipped in the package data.
    
    Examples of <ID>: Corsika7ID, Geant3ID, PythiaID.
```

src/particle/mcid.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class MCParticleID(int):
+    """
+    Base class for particle identification codes ("IDs") used by
+    Monte Carlo programs, convertible to and from the standard PDG IDs.
+
+    Concrete ID classes derive from this class and either
+
+    - define a class-level `_to_pdg_map` mapping of their IDs to PDG IDs,
```
eduardo-rodrigues:
```suggestion
    - Define a class-level `_to_pdg_map` mapping of their IDs to PDG IDs,
```

src/particle/mcid.py
```diff
@@ -0,0 +1,124 @@
...
+      in which case `to_pdgid` and `from_pdgid` are provided automatically, or
+    - override `to_pdgid` and `from_pdgid`, e.g. when the conversion
```
eduardo-rodrigues:
```suggestion
    - Override `to_pdgid` and `from_pdgid`, e.g. when the conversion
```

src/particle/mcid.py
```diff
@@ -0,0 +1,124 @@
...
+      is algorithmic and no explicit mapping is needed.
+
+    Examples
+    --------
+    Define a custom ID class given a known mapping to PDG IDs:
```
eduardo-rodrigues:
```suggestion
    Define a custom ID class `MyGeneratorID` given a known mapping to PDG IDs:
```

eduardo-rodrigues (commented):
This looks excellent! Thank you so much, @The-Ludwig. I left a few trivial things for you.

eduardo-rodrigues:
... I forgot to say: @henryiii, if you are too busy just let us know you will be skipping this review.

In any case this is the last PR before the major 1.0 release. DecayLanguage is awaiting this also for a 1.0 release :).

src/particle/corsika/corsika7id.py
The-Ludwig: Yeah, the distinction between non-existent and no corresponding ID was somewhat overlooked in the past, before Fable got to do it's magic.
The maps we generate are based on the ".csv" we generate with the admin script. The admin script just automatically creates all conversions between pdgids and corsika7 ids up to Iron-56: https://github.com/scikit-hep/particle/blob/80404c6c718d1b37c6daab98ea0f1364eac8b399/admin/dump_pdgid_to_corsika7.py#L160-L166

This is probably not optimal and can result in unexpected behaviour. So for 4818 in your example example,  it would never reach this point, since it is in the `pdg_map`. Iron-56 was an arbitrary choice, but is generally the heaviest element simulated with Corsika-7, as there are not really cosmic rays beyond that element. But in theory, there are both valid PDG and Corsika7 ID's for Oganesson, but then this logic here fails. The added check here mainly prevents non-sensical, but correct Corsika7 nuclides, like "z=0, a=1 == Neutron" and "z=2, a=2 == helium, but with no neutrons".

I think in this current state it is okay-ish, though we could re-run the admin script to include elements up to maybe A=250? That would make the csv's and internal maps considreably larger, but probably still on the super manageable level  (something below 1MB). But generally, I'd say this is sacrificing clean-ness and performance for very weird edge-cases, no user likely every runs into; If they do, I think this code is handable enough to dive into/send us a message

The-Ludwig:
Hi @eduardo-rodrigues !
Thank you so much for the review! I implemented your suggestions, the earlier overlook is a bit of a weird edge-case on which I commented.

src/particle/corsika/corsika7id.py
eduardo-rodrigues: OK, thank you for the extensive explanation. Would you mind add a bit of this to the admin script and/or the code to help anyone stumbling over the issue? Then we're good.

The-Ludwig: @eduardo-rodrigues I took 5 minutes reading the documentation for the CORSIKA7 output again ([here](https://www.iap.kit.edu/corsika/downloads/CORSIKA_GUIDE7.8050.pdf)), where on page 130 the particle numbering scheme is listed.
Actually, it explicitly says it does not allow `A > 56`, which is why the admin script stops there. I am not really sure why it shouldn't be able to go beyond 56, but apparently that's the documented description. 
It does not explicitly say Z>A is not allowed, which is why the code is structured in the way it is. (Although Z>A obviously doesn't make any sense whatsoever).  This means the code also didn't cover the edge case of A=59 and Z=99 before, which I now included. [If this ID ever appears anywhere, things have gone terribly wrong already... But the documentation does not explicitly forbid it.]

Since this is obviously completely confusing for anyone not working actively on Corsika7, I added quite extensive documentation about it with a link to the user guide of CORSIKA7. 
Since this is non-obvious to people not in the astroparticle-community: With a bit of luck this code never needs to be touched again, CORSIKA7 is mostly a legacy MC-program now, mainly being updated with bug-fixes by people in their free time. The C++ re-implementation CORSIKA8 just uses Pythia/PDGIDs. [Although adoption might still take 5-10 years until all the features and speed are en-par.]

admin/dump_pdgid_to_corsika7.py
```diff
@@ -156,8 +156,15 @@ def dump_pdgid_to_corsika7(file: pathlib.Path | None = None) -> None:
...
     (if file is None, else in the specified path).
     """
-    # Loop over all thinkable values and only add them if the PDG ID exists
+    # These are all nuclei allowed by CORSIKA7.
+    # Reference: Page 130 of the CORSIKA7 userguide Version 7.8050
+    # https://www.iap.kit.edu/corsika/downloads/CORSIKA_GUIDE7.8050.pdf
+    # Note: Some of these particles do not exist or make any sense.
+    # But in principle the CORSIKA7 numbering scheme allows them.
+    # CORSIKA7 stops at A=56, which is the number of nucleons
+    # in the most common iron-isotope
```
eduardo-rodrigues:
```suggestion
    # in the most common iron isotope.
```

admin/dump_pdgid_to_corsika7.py
```diff
@@ -156,8 +156,15 @@ def dump_pdgid_to_corsika7(file: pathlib.Path | None = None) -> None:
...
     for a in range(2, 56 + 1):
+        # `Particle.from_nucleus_info` throws an error anyway if Z>A
```
eduardo-rodrigues:
```suggestion
        # `Particle.from_nucleus_info` throws an error anyway if Z>A, which is not possible
```

src/particle/corsika/corsika7id.py
```diff
@@ -209,16 +199,19 @@ def to_pdgid(self) -> PDGID:
...
         """
-        from ..particle.particle import InvalidParticle  # pylint: disable=C0415
-
-        if self not in _bimap:
-            raise InvalidParticle(
-                f"The Corsika7ID {self} does not correspond to a particle and thus has no equivalent PDGID."
+        if int(self) in self._to_pdg_map:
+            return super().to_pdgid()
+
+        # Some nuclei with no known PDG ID are valid C7 IDs nonetheless
```
eduardo-rodrigues:
```suggestion
        # Some nuclei with no known PDG ID are valid C7 IDs nonetheless.
```

src/particle/corsika/corsika7id.py
```diff
@@ -209,16 +199,19 @@ def to_pdgid(self) -> PDGID:
...
+        # Particles with existing PDGIDs are in the csv conversion file
+        # and are caught by the first `if`
```
eduardo-rodrigues:
```suggestion
        # and are caught by the first `if`.
```

src/particle/corsika/corsika7id.py
```diff
@@ -209,16 +199,19 @@ def to_pdgid(self) -> PDGID:
...
+        # Reference: Page 130 of the CORSIKA7 userguide Version 7.8050
+        # https://www.iap.kit.edu/corsika/downloads/CORSIKA_GUIDE7.8050.pdf
```
eduardo-rodrigues:
```suggestion
        # https://www.iap.kit.edu/corsika/downloads/CORSIKA_GUIDE7.8050.pdf.
```

src/particle/corsika/corsika7id.py
eduardo-rodrigues: Thank you so much for this. As you guessed, I'm no user of Corsika and know nothing about it. We're good.

eduardo-rodrigues (approved):
Super to have this excellent enhancement going in. Thank you @The-Ludwig 🙏!

PR #799, opened 2026-09-10
fix(ci): give the publish job contents write permission for the zipapp release upload

### LINKED ISSUE #784: ZipApp not being deployed
See for example https://github.com/scikit-hep/particle/actions/runs/28178690149/job/83462087590.

@henryiii, I wonder if we want to fix this - it's been broken for a while, I see - or remove it.

henryiii:

The publish job's explicit `permissions` block dropped the default `contents: write`, so `softprops/action-gh-release` failed with "Resource not accessible by integration" when it attached `particle.pyz` to the release. This adds `contents: write` and makes `publish` depend on the `zipapp` job it downloads from.

Fixes #784

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -48,7 +48,7 @@ jobs:
           path: particle.pyz
 
   publish:
-    needs: [dist]
+    needs: [dist, zipapp]
     runs-on: ubuntu-latest
     if: github.event_name == 'release' && github.event.action == 'published'
     environment:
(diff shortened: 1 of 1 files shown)


eduardo-rodrigues (approved):
Hi @henryiii, thank you for the fix. I did not know it was just that keyword missing, unfortunately!

I hope the requirement on zipapp for the publishing won't be hitting us. Let's give it a try.

Thank you, merging.

